PR #1, opened 2017-07-07
Initial skeleton app

anatolykislov:


build.gradle
```diff
@@ -0,0 +1,27 @@
+buildscript {
```
sergey-surkov: Can you add copyright in all  source files in this form:

// Copyright 2017 Sourcerer Inc. All Rights Reserved.

resources/protos/commit.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto2";
```
sergey-surkov: Please use proto3

resources/protos/commit.proto
```diff
@@ -0,0 +1,22 @@
...
+
+option java_package = "app";
+option java_outer_classname = "CommitProtos";
+
+message Flag {
+    string type;
+    string value;
+}
+
+message Commit {
```
sergey-surkov: Can you document all protobuf messages and all their fields that are not obvious? Protobuf files are read often.

resources/protos/commit.proto
```diff
@@ -0,0 +1,22 @@
...
+    optional int32 id;
```
sergey-surkov: Does this protobuf compile at all? It should not because you must specify tags for all fields.

resources/protos/commit.proto
```diff
@@ -0,0 +1,22 @@
...
+    optional string email;
+
+    required string name = 1;
+    required int32 id = 2;
+    optional string email = 3;
+
+    repeated Flag flags;
```
sergey-surkov: What do flags do?

src/main/kotlin/app/Greeter.kt
```diff
@@ -0,0 +1,22 @@
+package app
+
+import app.utils.Args
+import app.utils.EmailValidator
+
+class Greeter(args: Args) {
+    var email: String? = args.email
+    var password: String? = args.password
+    var path: String? = args.path
```
sergey-surkov: This badly needs documentation. What is path? What is password?

src/main/kotlin/app/utils/EmailValidator.kt
```diff
@@ -0,0 +1,21 @@
...
+    @Throws(ParameterException::class)
+    override fun validate(name: String, value: String) {
+        if (!isValidEmail(value)) {
+            throw ParameterException("Parameter $name should be "
+                    + "correct email (found $value)")
+        }
+    }
+
+    fun isValidEmail(email: String): Boolean {
+        val ePattern = "^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@((\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\])|(([a-zA-Z\\-0-9]+\\.)+[a-zA-Z]{2,}))$"
```
sergey-surkov: Is there a way to bread down this line? Also, seriously we're using IP addresses for email hosts? I think we can safely skip that.

src/main/kotlin/app/utils/EmailValidator.kt
```diff
@@ -0,0 +1,21 @@
...
+        val p = java.util.regex.Pattern.compile(ePattern)
+        val m = p.matcher(email)
+        return m.matches()
+    }
+}
```
sergey-surkov: Please always add newline in the end of a file.

src/main/kotlin/app/utils/PathValidator.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import com.beust.jcommander.IParameterValidator
+import com.beust.jcommander.ParameterException
+
+object PathValidator : IParameterValidator {
+    @Throws(ParameterException::class)
+    override fun validate(name: String, value: String) {
+        if (!isValidPath(value)) {
+            throw ParameterException("Parameter $name should be "
+                    + "correct email (found $value)")
```
sergey-surkov: Copy-paste detected!

src/main/kotlin/app/utils/PathValidator.kt
```diff
@@ -0,0 +1,18 @@
...
+        }
+    }
+
+    fun isValidPath(email: String): Boolean {
+        return true
```
sergey-surkov: LOL

build.gradle
anatolykislov: Done

src/main/kotlin/app/utils/EmailValidator.kt
anatolykislov: Ok

resources/protos/commit.proto
```diff
@@ -1,22 +1,33 @@
...
 
 option java_package = "app";
 option java_outer_classname = "CommitProtos";
 
-message Flag {
-    string type;
-    string value;
+// Types of information that can be extracted from commits.
+enum EnumLabelTypes {
+    LOC_ADDED = 0; // Lines of code added.
```
sergey-surkov: Could you use two spaces before an inline comment? Pet peeve, but also if we start using Google code guides, it what they use across all languages.

resources/protos/commit.proto
```diff
@@ -1,22 +1,33 @@
...
-message Commit {
-    optional int32 id;
-    optional string email;
-
-    required string name = 1;
-    required int32 id = 2;
-    optional string email = 3;
+// Label is a specific extracted information about commit.
+message Label {
+    int32 type = 1; // Type of extracted information.
```
sergey-surkov: I think you are overgeneralizing. The approach you take makes it difficult working with this info. You will get something like "if type == loc_deleted do A, elif type == language do B". It's inefficient and does not serve any purpose. Please use dedicated fields for each attribute. E.g. 

```
message CommitStats {
  int32 num_lines_add = 1;
  int32 num_lines_deleted = 2;
  ...
}
```

When you use a structure like this your parsing code becomes non-existent.

resources/protos/commit.proto
```diff
@@ -1,22 +1,33 @@
...
+    string value = 2; // Extracted information.
+}
 
-    repeated Flag flags;
-}
\ No newline at end of file
+// Commit information message.
+message Commit {
+    int32 id = 1; // Commit id (hashed initial id).
+    int32 repo_id = 2; // Possibly hashed initial id of first commit.
```
sergey-surkov: Possibly? You can't write comments like that. What will happen if you'll forget it here, and then it will be a very mysterious comment. If you don't know how it's done, don't say it here. If you do know, make it clear what it is. Or you can put a todo:

```
int32 repo_id = 2;  // Repo id. TODO(anatoly): add details on how it is computed.
```

resources/protos/commit.proto
```diff
@@ -1,22 +1,33 @@
...
+    string author_name = 3;
+    string author_email = 4;
+    string date = 5;
+    repeated Label labels = 6; // List of labels.
```
sergey-surkov: So this list of labels, it not really a list of labels, right? it is stats on a commit. Here is a list of labels: "apples", "pears", "blueberries"

src/main/kotlin/app/Greeter.kt
```diff
@@ -1,13 +1,15 @@
...
 
 import app.utils.Args
 import app.utils.EmailValidator
 
 class Greeter(args: Args) {
-    var email: String? = args.email
-    var password: String? = args.password
-    var path: String? = args.path
-    var silent: Boolean = args.silent
+    var email: String? = args.email // Sourcerer account email.
```
sergey-surkov: I don't believe we'll be using an email as user id. We will be giving them handles just like github does.

src/main/kotlin/app/utils/PasswordValidator.kt
```diff
@@ -12,7 +14,7 @@ object PasswordValidator : IParameterValidator {
         }
     }
 
-    fun isValidPassword(email: String): Boolean {
-        return true
+    fun isValidPassword(password: String): Boolean {
+        return true //TODO Implement validation
```
sergey-surkov:
```
// TODO(anatoly): Implement validation.
```

src/main/kotlin/app/utils/PathValidator.kt
```diff
@@ -8,11 +10,11 @@ object PathValidator : IParameterValidator {
...
             throw ParameterException("Parameter $name should be "
-                    + "correct email (found $value)")
+                    + "correct path (found $value)")
         }
     }
 
-    fun isValidPath(email: String): Boolean {
-        return true
+    fun isValidPath(path: String): Boolean {
+        return true //TODO Implement validation
```
sergey-surkov: ditto

src/main/kotlin/app/utils/PathValidator.kt
```diff
@@ -8,11 +10,11 @@ object PathValidator : IParameterValidator {
     override fun validate(name: String, value: String) {
         if (!isValidPath(value)) {
             throw ParameterException("Parameter $name should be "
-                    + "correct email (found $value)")
+                    + "correct path (found $value)")
```
sergey-surkov: It will probably be more readable if you break the line right after ParameterException:

```
throw ParameterException(
    "Parameter $name should be a correct path (found $value)")
```

src/main/proto/commit.proto
```diff
@@ -7,27 +7,22 @@ package app;
...
-    int32 repo_id = 2; // Possibly hashed initial id of first commit.
+    int32 id = 1;  // Commit id. Hashed initial id.
+    int32 repo_id = 2;  // Repo id. TODO(anatoly): add details on how it is
+        // computed.
     string author_name = 3;
     string author_email = 4;
     string date = 5;
-    repeated Label labels = 6; // List of labels.
+    bool qommit = 6;  // Quality commit.
+    repeated CommitStats commit_stats = 7;  // Stats on a commit.
```
sergey-surkov: it's stats inside a commit. just call it 'stats'.

src/main/proto/commit.proto
```diff
@@ -24,5 +24,7 @@ message Commit {
     string author_email = 4;
     string date = 5;
     bool qommit = 6;  // Quality commit.
-    repeated CommitStats commit_stats = 7;  // Stats on a commit.
+    int32 num_lines_add = 7;  // Number of lines for whole commit.
```
sergey-surkov: These are new and they are not obviously useful. Presumably you can compute them by aggregating Stats. Usually, when you respond to comments in code review, you should refrain from adding unexpected non-obvious changes (i.e. changes that don't pertain to review comments), because this will inevitably delay the process. If you HAVE to add those, please explain why in your commit comment.

I am going to approve this now, but please remove these new fields, and include them in a future commit if they are actually needed.

Don't forget to squash the dev branch before pushing. Also - should we merge or rebase? For now we should probably do rebase because it ends up in a cleaner master.

anatolykislov: Ok, will do it. I think that this fields necessary because if we can't recognize and calculate stats for all of the lines (unknown format for example) we will get the wrong aggregation.
I agree with you about rebase for squashing.

sergey-surkov: Makes sense. In such a case, it makes sense to make their purpose obvious by provide a detailed comment for this field in the proto file.

PR #2, opened 2017-07-13
Configurator and updated workflow

anatolykislov:
Updated application workflow according to convenient command-line applications (that the application was more like a git).

Usage: sourcerer [options] [command] [command options]
  Options:
    -u, --username
      Sourcerer account username
    -p, --password
      Sourcerer account password
    -s, --silent
      Silent mode
  Commands:
    explore      Explore statistics on specific repository
      Usage: explore REPOPATH
    config      Configure Sourcerer app
      Usage: config [options] KEY VALUE
        Options:
          --local
            Used to specify local level of configuration (default)
          --user
            Used to specify user level of configuration

Class Configurator is finished. Command config can be used.

Added number of added and deleted lines to Commit proto message because aggregated number from Stats may be not accurate.


src/main/kotlin/app/Configurator.kt
```diff
@@ -0,0 +1,199 @@
+// Copyright 2017 Sourcerer Inc. All Rights Reserved.
+
+package app
+
+import app.utils.Options
+import app.utils.PasswordValidator
+import app.utils.UsernameValidator
+import org.apache.commons.configuration.ConfigurationException
```
anatolykislov: [3.3.3 Ordering and spacing](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) says that import statements have to be sorted in ascii sort order.

src/main/kotlin/app/Configurator.kt
```diff
@@ -0,0 +1,178 @@
...
+
+        if (pair.count() != 2) {
+            return options
+        }
+
+        val (key, value) = pair
+
+        when (key) {
+            configUsername -> {
+                if (UsernameValidator().isValidUsername(value)) {
```
sergey-surkov: I am confused. You already connect validators to options in Options class. Shouldn't they get invoked automatically on assignment?

sergey-surkov: Also, this seems wrong. Why do you need to manually assign options here from essentially what amounts a command line? Should there be a built-in mechanism for it? Why does this function exist?

src/main/kotlin/app/Explorer.kt
```diff
@@ -0,0 +1,30 @@
+// Copyright 2017 Sourcerer Inc. All Rights Reserved.
+// Author: Anatoly Kislov (anatoly@sourcerer.io)
+
+package app
+
+import app.utils.CommandExplore
+import java.io.File
+
+/**
+ * Explorer is a class that does repository analysis and sends statistics data
```
anatolykislov: Ok

src/main/kotlin/app/Explorer.kt
```diff
@@ -0,0 +1,30 @@
...
+ * to Sourcerer.
+ */
+class Explorer(options: CommandExplore) {
```
anatolykislov: Ok

src/main/kotlin/app/Main.kt
```diff
@@ -1,25 +1,59 @@
...
+        else -> help(jc)  // Show help info if no command specified.
     }
 
-    println(arrayOf(args.username, args.password, args.path).joinToString())
+    println(arrayOf(options.username, options.password).joinToString())
+    println(jc.parsedCommand)
+    println(explore.path)
+}
+
+fun explore(commandOptions: CommandExplore) {
```
anatolykislov: Shouldn't it be just: add, remove, list? Is there anything else that may have the same set of commands?

src/main/kotlin/app/Configurator.kt
```diff
@@ -0,0 +1,178 @@
...
+        current = options
+    }
+
+    fun loadUserLevelConfig() {
+        if (userDir != null) {
+            config = loadConfig(userDir)
+        }
+    }
+
+    fun saveUserLevelConfig(options: Options): Boolean {
```
anatolykislov: I'll rename the function, but we need it still because it's being called from Main when config command is used.

src/main/kotlin/app/Configurator.kt
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        return false
+    }
+
+    fun setCurrentOptions(options: Options) {
+        current = options
+    }
+
+    fun loadUserLevelConfig() {
```
anatolykislov: Ok

src/main/kotlin/app/Configurator.kt
anatolykislov: While using app one could specify a username, password, etc as global options before a command. See usage. But if one wants to update a configuration he should use command config and its options (key value pair). Validators invoke only on specific parameters not on abstract key value pair, so that's why I validate it manually here. 

Usage: sourcerer [options] [command] [command options]
  Options:
    -u, --username
      Sourcerer account username
    -p, --password
      Sourcerer account password
    -s, --silent
      Silent mode
  Commands:
    explore      Explore statistics on specific repository
      Usage: explore REPOPATH

    config      Configure Sourcerer app
      Usage: config KEY VALUE

src/main/kotlin/app/utils/PasswordValidator.kt
```diff
@@ -15,6 +16,6 @@ object PasswordValidator : IParameterValidator {
     }
 
     fun isValidPassword(password: String): Boolean {
-        return Regex("^[a-zA-Z0-9_.+-]$").containsMatchIn(password)
+        return Regex("^[a-zA-Z0-9_.+-]+$").containsMatchIn(password)
```
anatolykislov: Sounds good to me. So should we use sha hash on server?

src/main/kotlin/app/Main.kt
sergey-surkov: I guess add/remove/list will work. Also, run should probably ask login and repo info interactively, if it was not previously configured.

src/main/kotlin/app/utils/PasswordValidator.kt
sergey-surkov: Yeah, let's store password hashed, and have the server to expect such a hash.

src/main/kotlin/app/Configurator.kt
sergey-surkov: I really think we should have @osilla define user UI. Ideally, I would want them to launch sourcerer, get asked what their user info is, etc, and have that info saved to a config file. If they ever need to change it, they can always delete it from the config file. My worry here is that by building a complicated options/commands system, we will make it necessary to read documentation before they can use it. Of course, we can always provide a wrapper bash for this, so maybe it's the way to go, but I would like @osilla to have authority on UI because it's how we make UI consistent.

src/main/kotlin/app/Configurator.kt
sergey-surkov: My point was it does not do enough to warrant a separate function. You can make its buddy into the only place where it's being called, and you will end up with less code, and an easier to read codebase.

anatolykislov: Ok

src/main/kotlin/app/Configurator.kt
anatolykislov: I think that it's more convenient to use the application on options/commands way but maybe you are right about documentation and we could make initial execution be more like a tutorial.

src/main/kotlin/app/Configurator.kt
```diff
@@ -0,0 +1,178 @@
...
+import org.apache.commons.configuration.PropertiesConfiguration
+
+/**
+ * Configurator is a singleton class that manages configuration files and
+ * values of non-command specific options.
+ */
+object Configurator {
+    val configFileName = "sourcerer.properties"
+    val configUsername = "username"
+    val configPassword = "password"
```
anatolykislov: Ok

src/main/kotlin/app/Configurator.kt
```diff
@@ -0,0 +1,178 @@
...
+        loadUserLevelConfig()
+    }
+
+    // Merges different levels of options into one options object in the next
+    // order: current, local, user, default.
+    private fun mergeLevels(): Options {
+        val levels = arrayListOf(current, config, default)
+        val merged = Options()
+
+        for (level in levels) {
```
anatolykislov: It could be done with reflection. But it will increase the overall size of the app and can have some performance drawback.

src/main/kotlin/app/RepoExplorer.kt
```diff
@@ -0,0 +1,29 @@
...
+    fun explore() {
+        var isValidPath = false
+
+        try {
+            val f = File(path)
+            isValidPath = f.isDirectory
+        } catch (e: SecurityException) {
+        }
+
+        println(isValidPath)
```
yandzee: it would be better if you use logger,
check this https://github.com/jupf/staticlog

anatolykislov: Ok

sergey-surkov:
Guys what is going on? The request is approved, why do you continue piling stuff on top of it?

PR #3, opened 2017-07-15
Install

asurkov:


README.md
```diff
@@ -1,2 +1,17 @@
 # sourcerer-app
 Sourcerer app makes engineering profiles from their git repositories.
```
sergey-surkov: We are experimenting with language. Can you please change it to:

Sourcerer hashes your git repos into intelligent engineering profiles.

README.md
```diff
@@ -1,2 +1,17 @@
...
+
+Install/uninstall
+=================
+
+Run the command to install `sourcerer` tool on your machine:
```
sergey-surkov: To install sourcerer run the following command:

README.md
```diff
@@ -1,2 +1,17 @@
...
+
+```
+curl -s https://sourcerer.io/install | sudo bash
+```
+
+Run the command to remove it from your machine:
```
sergey-surkov: To remove sourcerer from your machine:

README.md
```diff
@@ -1,2 +1,17 @@
...
+
+```
+curl -s https://sourcerer.io/uninstall | sudo bash
```
sergey-surkov: This worries me. Sourcerer will inevitably change, so will the uninstall script. And the version matching will become tricky. I think you should be installing the uninstall script locally during the installation.

src/install/install
```diff
@@ -0,0 +1,20 @@
+echo "Installing sourcerer app.."
```
sergey-surkov: Add copyright. We will revise this once we agree on the license for this.
```
# Copyright 2017 Sourcerer, Inc. All Rights Reserved.
#

```

src/install/install
```diff
@@ -0,0 +1,20 @@
...
+
+if [ -f /usr/local/bin/sourcerer ] ; then
+  read -e -p "Previous sourcerer app installation is detected. Reinstall it? [Y/n] " yesno < /dev/tty
+  if [ "$yesno" ] && [ "$yesno" = "n" ] ; then
+    echo "Exit.."
```
sergey-surkov: Exiting

src/install/install
```diff
@@ -0,0 +1,20 @@
+echo "Installing sourcerer app.."
+
+if [ -f /usr/local/bin/sourcerer ] ; then
+  read -e -p "Previous sourcerer app installation is detected. Reinstall it? [Y/n] " yesno < /dev/tty
```
sergey-surkov: Previous version of sourcerer is detected. Reinstall it? [Y/n]

src/install/install
```diff
@@ -0,0 +1,20 @@
...
+  if [ "$yesno" ] && [ "$yesno" = "n" ] ; then
+    echo "Exit.."
+    exit
+  fi
+fi
+
+curl -s https://sourcerer.io/download > /usr/local/lib/sourcerer.jar
```
sergey-surkov: I strongly recommend to parametrize the urls.

e.g. 
```
SERVER='https://sourcerer.io'
DOWNLOAD_URL=$SERVER/download
INSTALL_DIR=/usr/local/bin
```
etc.

src/install/install
```diff
@@ -0,0 +1,20 @@
...
+echo "java -jar /usr/local/lib/sourcerer.jar" > /usr/local/bin/sourcerer
+chmod +x /usr/local/bin/sourcerer
+
+echo "Done"
```
sergey-surkov: echo "Done!"
echo "Run sourcerer to start hashing your repos!"

src/install/uninstall
```diff
@@ -0,0 +1,10 @@
+read -e -p "The script will uninstall sourcerer app. Proceed? [Y/n] " yesno < /dev/tty
```
sergey-surkov: Copyright, etc

README.md
asurkov: np, I see you like fix-while-you're-around approach, which is fine for sure

README.md
asurkov: Do you foresee we will go beyond the single sourcerer.jar? Otherwise this is a safe assumption I'd say. On the other hand a server side uninstall script could remove everything it installed, ever. This may be a reasonable approach, if the install script is more or less stable and not changed all the time.

Anyway, if you really prefer a local uninstall script, then it'd be a nice option to have something like:
sourcerer --uninstall
but not sure if the script can remove itself while running. What's your thinking?

sergey-surkov: Uninstallation is not just a jar. There is also a config file, there can be some sort of caching files. Maybe temp files. Formats of those will change, we need to be careful. So yes, I would rather have an uninstall script locally. In Unix, a program can remove itself, but in Windows last time I checked it was not the case. I would prefer a separate script.

README.md
sergey-surkov: We need consistent messaging. The 'hashing' language is what we decided to adopt across the product.

README.md
asurkov: I've got used to think (in OS X world) uninstallation process is that simple as removing the app itself (which is a folder technically). In Linux world people do apt-get remove. It seems neither of these options works in this case. So what name and location for this script could be?

README.md
asurkov: I don't argue about the wording. Just some people prefer to have separate issue for not related changes. It's more about style, which differs from company to company, from team to team, etc

README.md
asurkov: I confirm that I didn't miss this comment and I think it is reasonable.

README.md
asurkov: I confirm that I didn't miss this comment and I think it is reasonable.

src/install/install
asurkov: I confirm that I didn't miss this comment and I think it is reasonable.

src/install/install
asurkov: Exiting wihout .. or with ..? Otherwise I confirm that I didn't miss this comment and I think it is reasonable.

src/install/install
asurkov: I confirm that I didn't miss this comment and I think it is reasonable.

src/install/uninstall
asurkov: I confirm that I didn't miss this comment and I think it is reasonable. Not yet sure, why I am doing this :)

src/install/install
asurkov: Aha! I missed this one to comment out :) Confirming it's already fixed.

PR #4, opened 2017-07-17
Add build info to README APP-12

anatolykislov:


README.md
```diff
@@ -4,6 +4,8 @@ Sourcerer hashes your git repos into intelligent engineering profiles.
 Install/uninstall
 =================
 
+Sourcerer requires JRE to be installed on the system. Install last version from [Oracle](http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html).
+
```
asurkov: as I mentioned already, it wasn't enough in my case, JDK was required. We might want to file a follow up issue to investigate it.

README.md
```diff
@@ -15,3 +17,22 @@ To remove sourcerer from your machine:
 ```
 sudo sourcerer --uninstall
 ```
+
+Build
+=====
+
+To build and run this application, you'll need latest versions of Git, Gradle and JDK installed on your computer. From your command line:
+
```
asurkov: I would move this into Prerequisites section that will explain the user how to prepare his machine. Otherwise it looks good with me.

README.md
anatolykislov: JRE should be enough if you only want to run the JAR file that you downloaded. If you want to build the JAR and run it from source code then you have to install JDK.
If you do have troubles with running the JAR file without JDK (only with JRE) we should definitely investigate it.

sergey-surkov:
Since Alex had concerns, he should be the one approving this.

asurkov:
I don't have any the patch-stopper concerns. My comments can be addressed in follow ups. I didn't thumb up it myself, leaving it to @sergey-surkov, mainly to check the language/doc structure since it's user-facing component.

PR #6, opened 2017-07-20
APP-17 Installation script should report which Java Platform is requi…

asurkov:
It also moves "Please also install JDK on your machine" message from the installation script itself under 'sourcerer' script which will notify the user about missed 'java' tool every time the script is running.


README.md
```diff
@@ -4,7 +4,9 @@ Sourcerer hashes your git repos into intelligent engineering profiles.
 Install/uninstall
 =================
 
-Sourcerer requires JRE to be installed on the system. Install last version from [Oracle](http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html).
+Sourcerer requires Java Platform installed on the system.
+* Proceed with JRE(http://www.oracle.com/technetwork/java/javase/downloads/jre8-downloads-2133155.html) if on Linux,
```
sergey-surkov: Why not JRE on all platforms?

asurkov: It seems there's something mac specific. I  was required to install JDK after I installed JRE. @anatolykislov I believe observed same thing.

asurkov: here's more info [1]:

Using the Command-line to find Java Versions - Mac

The Java Runtime (JRE) that you download from java.com or oracle.com contains a plugin to run Java content from your browser. In order to use the command line tools, you will need to download the Java Development Kit (JDK). The JRE and JDK are separate and can coexist on your system. Only one JRE can be installed on Mac. There can be multiple JDKs installed on a system, as many as you wish.

[1] https://www.java.com/en/download/help/version_manual.xml#cmdline

sergey-surkov: Damn you Oracle.

asurkov: :)
so many obstacles on the way :)

anatolykislov: Agree on this.

PR #8, opened 2017-07-27
feat: add basic repo processing, refactor SourcererApi, refactor AuthState, fix Jackson serialization (APP-4, APP-7)

anatolykislov:
Fix:
- Getter serialization in Jackson disabled for ProtoWrapper classes (tried to serialize proto because of getProto)

Repo processing:
- Base implementation of reading commits from local repo and sending commit messages to server
- Add JGit library
- RepoExplorer renamed to RepoHasher
- Add constructor from JGit's commit to Commit class

Network:
- Refactor of SourcererApi
- ProtoWrapper now serializes to bytes
- Add RequestException wrapper for FuelError
- Refactor of AuthState
- Add auth and connection errors processing in AuthState


src/main/kotlin/app/RepoHasher.kt
```diff
@@ -0,0 +1,89 @@
...
+import app.utils.RepoHelper
+import io.reactivex.Observable
+import org.eclipse.jgit.api.Git
+import org.eclipse.jgit.revwalk.RevCommit
+import org.eclipse.jgit.revwalk.RevWalk
+import java.io.File
+import java.io.IOException
+
+/**
+ * RepoHasher hashes repository and uploads stats to server.
```
asurkov: it's good to extend the comment by saying what repository hashing exactly mean

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -0,0 +1,89 @@
...
+        //TODO(anatoly): Implement data transfer.
+
+        var firstCommit: RevCommit
+
+        // Get commits loading observable and wait for subscribers.
+        val observableRevCommits = getObservableRevCommits().publish()
+
+        // Main data flow. Read commits from repo and send them to server.
+        observableRevCommits
+                .doOnNext {
```
asurkov: styling issue: what the space offset do you use? it seems there's inconsistency between a function body offset and this one

src/main/kotlin/app/utils/RequestException.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    constructor(fuelError: FuelError) : this(fuelError as Exception) {
+        httpStatusCode = fuelError.response.httpStatusCode
+        httpResponseMessage = fuelError.response.httpResponseMessage
+    }
+
+    fun isAuthError(): Boolean {
+        return httpStatusCode == 403
+    }
+}
```
asurkov: nit: no new line

src/main/kotlin/app/ui/AuthState.kt
```diff
@@ -6,51 +6,80 @@ package app.ui
...
+        }
+    }
 
-                break
+    fun tryAuth(): Boolean {
+        try {
+            doAuth()
+            return true
+        } catch (e: RequestException) {
+            if(e.isAuthError()) {
```
asurkov: nit: space after if

src/main/kotlin/app/ui/AuthState.kt
```diff
@@ -6,51 +6,80 @@ package app.ui
...
-
-            isCredentialsUnknown = true
         }
+        return false
     }
 
-    override fun next() {
-        context.changeState(ListRepoState(context))
+    fun doAuth() {
+        println("Authorising...")
```
asurkov: technically doAuth is just a body of a try block inside tryAuth function, I'd say it makes sense to merge these two functions

src/main/kotlin/app/SourcererApi.kt
```diff
@@ -25,74 +26,70 @@ object SourcererApi {
...
+        Logger.debug("Request $requestName success")
+    }
+
+    fun getUserBlocking(): User {
+        return makeBlockingRequest(createRequestGetUser(),
+                "getUserBlocking",
+                { body -> User().parseFrom(body) })
+    }
+
+    fun getUserAsync(success: (String) -> Unit, failure: (String) -> Unit) {
```
asurkov: naming: I would probably name those as getUser and getUserAsync.

Also I'm curious if the API really needs to support both types of calls: sync and async.

src/main/kotlin/app/RepoHasher.kt
anatolykislov: We using the word "to hash" as an analog of "to analyse" in app description and in app UI. I guess to explain what does it mean in source code comments is redundant.

src/main/kotlin/app/RepoHasher.kt
anatolykislov: +4 indent
+8 line wrap

src/main/kotlin/app/utils/RequestException.kt
anatolykislov: Ok, will fix

src/main/kotlin/app/ui/AuthState.kt
anatolykislov: Ok, will fix

src/main/kotlin/app/ui/AuthState.kt
anatolykislov: Ok

src/main/kotlin/app/SourcererApi.kt
anatolykislov: The reason here is to make a more clear description of the behaviour of functions.
For example, getUserBlocking used in UI while blocking auth check. postCommitAsync used in repo hashing to send prepared data while processing remaining commits.

src/main/kotlin/app/RepoHasher.kt
yandzee: In many langs first method call in chain places at the same line as receiver, so i would write `observableRevCommits.doOnNext {`

src/main/kotlin/app/Configurator.kt
```diff
@@ -164,16 +177,13 @@ object Configurator {
...
-        mapper.registerModule(KotlinModule()) // Enable Kotlin support.
-
         // Сonfig initialization in case an exception is thrown.
         var loadConfig = Config()
 
         try {
             loadConfig = Files.newBufferedReader(Paths.get(userDir,
                     CONFIG_FILE_NAME)).use {
-                mapper.readValue(it, Config::class.java)
+                createMapper().readValue(it, Config::class.java)
```
sergey-surkov: Is this mapper creation thing expensive? If it is, you could create it once and store it in a field.

src/main/kotlin/app/RepoHasher.kt
sergey-surkov: Yeah, it's the language thing we agreed up. We say that 'Sourcerer hashes your repos into a profile' to stress the fact that we don't upload code anywhere.

src/main/kotlin/app/RepoHasher.kt
sergey-surkov: +1 to what @yandzee sez. Keep it all on the same line.

sergey-surkov: Also 8-space line wrap, is it our invention or Kotlin's? Seems like a lot given our 80 char rule.

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -0,0 +1,89 @@
...
+                    Logger.info("Commit: ${it.id.name} ${it.shortMessage}")
+                }
+                .map { c -> Commit(c) }
+                .subscribe(
+                        {
+                            Logger.debug("onNext")
+                            SourcererApi.postCommitAsync(Commit(), {}, {})
```
sergey-surkov: Why are you submitting an empty commit?

Shouldn't there be a commit data structure that listers fill out? Also, is there a guarantee this will be called after all other listeners were called?

src/main/kotlin/app/RepoHasher.kt
anatolykislov: Ok, will fix

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -0,0 +1,89 @@
...
+
+        // Get commits loading observable and wait for subscribers.
+        val observableRevCommits = getObservableRevCommits().publish()
+
+        // Main data flow. Read commits from repo and send them to server.
+        observableRevCommits
+                .doOnNext {
+                    Logger.info("Commit: ${it.id.name} ${it.shortMessage}")
+                }
+                .map { c -> Commit(c) }
```
sergey-surkov: I think this chaining is confusing here. At the very least a subscription seem to belong a separate call. Chaining is useful when you can think of individual operations as building on top of each other. Here, a subscription is arbitrary, and will follow up by other subscribers (presumably).

src/main/kotlin/app/SourcererApi.kt
yandzee: Async/blocking is not a property of certain functionality but of entire architecture; its strange moment

src/main/kotlin/app/SourcererApi.kt
```diff
@@ -25,74 +26,69 @@ object SourcererApi {
...
+        try {
+            Logger.debug("Request $requestName initialized")
+            val (_, _, result) = request.responseString()
+            val body = result.get()
+            Logger.debug("Request $requestName success")
+            return parser(body)
+        } catch (e: FuelError) {
+            Logger.error("Request $requestName error", e)
+            throw RequestException(e)
+        }
```
yandzee: There is no need to try { .. }; FuelError is a second element in result.

src/main/kotlin/app/SourcererApi.kt
```diff
@@ -25,74 +26,69 @@ object SourcererApi {
...
-        val (request, response, result) = Fuel.get("/commit")
-                .authenticate(username, password)
+    private fun createRequestPostCommit(commit: Commit): Request {
+        return Fuel.post("/commit").authenticate(username, password)
                 .body(commit.serialize())
-                .responseString()
+    }
+
+    private fun <T> makeBlockingRequest(request: Request, requestName: String,
+                                parser: (String) -> T): T {
```
sergey-surkov: I would recommend to align arguments under each other. Will be more readable. E.g.
```
private fun<T> makeBlockingRequest(request: Request, requestName: String,
                                   parser: (String) -> T): T {
```

src/main/kotlin/app/SourcererApi.kt
```diff
@@ -25,74 +26,69 @@ object SourcererApi {
...
+                                 failure: (String) -> Unit) {
+        Logger.debug("Request $requestName initialized")
+        request.responseString { _, _, result ->
+            when (result) {
+                is Result.Failure -> {
+                    Logger.error("Request $requestName error",
+                            result.getException())
+                    failure(result.get())
+                }
+                is Result.Success -> {
```
sergey-surkov: I would prefer to see Success go before Failure. Happy path is a main path.

src/main/kotlin/app/SourcererApi.kt
```diff
@@ -25,74 +26,69 @@ object SourcererApi {
...
+                    success(result.get())
                 }
+            }
+        }
+        Logger.debug("Request $requestName success")
+    }
+
+    fun getUserBlocking(): User {
+        return makeBlockingRequest(createRequestGetUser(),
+                "getUserBlocking",
```
sergey-surkov: yeah, aligning this under args would look good

src/main/kotlin/app/SourcererApi.kt
```diff
@@ -25,74 +26,69 @@ object SourcererApi {
...
+    }
+
+    fun getUserAsync(success: (String) -> Unit, failure: (String) -> Unit) {
+        makeAsyncRequest(createRequestGetUser(), "getUserAsync",
+                success, failure)
+    }
+
+    fun postCommitBlocking(commit: Commit): String {
+        return makeBlockingRequest(createRequestPostCommit(commit),
+                "postCommitBlocking", {body -> body})
```
yandzee: { it }

src/main/kotlin/app/SourcererApi.kt
sergey-surkov: +1 to unclear we need both. I would only add another version if you use both. You are not writing a library, you are writing a program. Any code that does not get called is dead code.

src/main/kotlin/app/model/Commit.kt
```diff
@@ -20,6 +22,13 @@ class Commit : ProtoWrapper<Commit, CommitProtos.Commit> {
     var numLinesDelete: Int = 0
```
sergey-surkov: s/numLinesDelete/numLinesDeleted/

src/main/kotlin/app/model/Commit.kt
```diff
@@ -20,6 +22,13 @@ class Commit : ProtoWrapper<Commit, CommitProtos.Commit> {
...
     // TODO(anatoly): add Stats.
 
+    constructor(revCommit: RevCommit) : this() {
+        this.id = DigestUtils.sha256Hex(revCommit.id.name)
+        this.author = Author(revCommit.authorIdent.name,
+                revCommit.authorIdent.emailAddress)
```
sergey-surkov: yeah.. alignment..

src/main/kotlin/app/model/Commit.kt
```diff
@@ -20,6 +22,13 @@ class Commit : ProtoWrapper<Commit, CommitProtos.Commit> {
...
+        this.dateTimestamp = revCommit.commitTime
```
sergey-surkov: What did we decide on versioning of commits? Because this would be a good place to register the time processing the commit.

src/main/kotlin/app/model/ProtoWrapper.kt
```diff
@@ -11,15 +11,15 @@ import com.google.protobuf.GeneratedMessageV3
 interface ProtoWrapper<out T, P : GeneratedMessageV3> {
     fun getProto(): P
 
-    fun serialize(): String {
-        return this.getProto().toByteArray().toString(Charsets.UTF_8)
+    fun serialize(): ByteArray {
+        return this.getProto().toByteArray()
```
sergey-surkov: Lol, so maybe this class is not needed, after all.

src/main/kotlin/app/ui/AuthState.kt
```diff
@@ -6,51 +6,76 @@ package app.ui
...
+    }
 
-                // Save username and password to config file.
-                if (isCredentialsUnknown) {
-                    Configurator.setUsernamePersistent(username)
-                    Configurator.setPasswordPersistent(password)
-                    Configurator.saveToFile()
-                }
+    fun getUsername() {
+        println("Enter Sourcerer account username:")
```
sergey-surkov: You could just say
username:
like they all do in unix

src/main/kotlin/app/ui/AuthState.kt
```diff
@@ -6,51 +6,76 @@ package app.ui
...
+        username = readLine() ?: ""
+        Configurator.setUsernameCurrent(username)
+    }
 
-                break
-            }
+    fun getPassword() {
+        println("Enter Sourcerer account password:")
```
sergey-surkov: just password

src/main/kotlin/app/ui/AuthState.kt
```diff
@@ -6,51 +6,76 @@ package app.ui
...
+        if (username.isNotEmpty() || password.isNotEmpty()) {
+            Configurator.saveToFile()
         }
     }
 
-    override fun next() {
-        context.changeState(ListRepoState(context))
+    fun tryAuth(): Boolean {
+        try {
+            println("Authorising...")
```
sergey-surkov: I think you are authenticating here, not authorizing

src/main/kotlin/app/ui/UpdateRepoState.kt
```diff
@@ -3,13 +3,18 @@
...
  * Update repositories console UI state.
  */
 class UpdateRepoState constructor(val context: Context) : ConsoleState {
     override fun doAction() {
-        // TODO(anatoly): Implement.
         println("Hashing your git repositories.")
+        for (repo in Configurator.getRepos()) {
+            RepoHasher(repo).update()
+        }
         println("The repositories has been hashed. See result online on your "
```
sergey-surkov: have been

src/main/kotlin/app/utils/RequestException.kt
```diff
@@ -0,0 +1,21 @@
+// Copyright 2017 Sourcerer Inc. All Rights Reserved.
+// Author: Anatoly Kislov (anatoly@sourcerer.io)
+
+package app.utils
+
+import com.github.kittinunf.fuel.core.FuelError
+
+class RequestException(val exception: Exception)
+    : Exception(exception.message) {
```
sergey-surkov: It's weird it's all aligned with code. shouldn't you do your 8 spaces here? Or why not keep it on the same line?

src/main/kotlin/app/SourcererApi.kt
anatolykislov: Ok, it's better

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -0,0 +1,89 @@
...
+        } catch (e: IOException) {
+            subscriber.onError(e)
+            null
+        }
+
+        if (git != null) {
+            val repository = git.repository
+            try {
+                val revWalk = RevWalk(repository)
+                val commitId = repository.resolve("refs/heads/master")
```
sergey-surkov: This is a very much a shotgun approach. We need to be careful here, and only take commits that belong to a particular user. We should look into repo/.git/config and ~/.gitconfig for user's emails in those repos, and use those to pick commits. Also, later on, there will be a need to generate profiles for user's colleagues (i.e. people the user actively works with), so this needs to be organized in a way that one iteration over commits targets a specific user identified by a set of email addresses.

At the very least, add a TODO.

src/main/kotlin/app/SourcererApi.kt
anatolykislov: I know. try/catch is more common and better to understand way.

src/main/kotlin/app/Configurator.kt
anatolykislov: Ok

src/main/kotlin/app/RepoHasher.kt
anatolykislov: It's a mistake. Will fix.
It is definitely will be called after all chain links.

src/main/kotlin/app/SourcererApi.kt
anatolykislov: Ok

src/main/kotlin/app/SourcererApi.kt
anatolykislov: Ok

src/main/kotlin/app/SourcererApi.kt
anatolykislov: Ok

src/main/kotlin/app/model/Commit.kt
anatolykislov: We could save the time of processing on the server when data saved to DB. No need to send additional data.

src/main/kotlin/app/ui/AuthState.kt
anatolykislov: Ok

src/main/kotlin/app/ui/AuthState.kt
anatolykislov: Ok

src/main/kotlin/app/ui/AuthState.kt
anatolykislov: You are right

src/main/kotlin/app/ui/UpdateRepoState.kt
anatolykislov: Ok

src/main/kotlin/app/utils/RequestException.kt
anatolykislov: Ok, I'll fix. Weird behavior of IDE.
It doesn't fit the 80.

src/main/kotlin/app/RepoHasher.kt
anatolykislov: Ok, will add a TODO.

src/main/kotlin/app/RepoHasher.kt
asurkov: My point was that "RepoHasher hashes repository" is not that descriptive as it could be, I would prefer to use 'analyze' word if it's indeed a synonym of 'hashes', something like "RepoHasher analyzes a repository, extract the stats and uploads it to the server"

src/main/kotlin/app/model/ProtoWrapper.kt
anatolykislov: Maybe but not now.

src/main/kotlin/app/model/ProtoWrapper.kt
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 Sourcerer Inc. All Rights Reserved.
+// Copyright 2017 Sourcerer  . All Rights Reserved.
```
sergey-surkov: What happened here??

anatolykislov: Fixed

PR #10, opened 2017-08-04
wip: edit proto file, add Stats and Commits protobuf wrappers, remove…

anatolykislov:
… ProtoWrapper interface (APP-22)

- proto file finished
- add Stats and Commits protobuf wrappers
- removed ProtoWrapper interface, after removing convertion to string with specified charset from last PR no actual need in this interface
- add primary constructors to all protobuf wrapper (see [named args](https://kotlinlang.org/docs/reference/functions.html)), converted parseFrom methods to secondary constructors


src/main/proto/sourcerer.proto
```diff
@@ -69,7 +69,7 @@ message Repo {
     string initial_commit_rehash = 2;
 
     // Authors' email filter for hashed commits.
-    bool email_filter_enabled = 3;
+    bool is_email_filter_enabled = 3;
```
sergey-surkov: please kill it as discussed offline

PR #12, opened 2017-08-13
feat: add stats extractors, add commit diff calculation, add contribu…

anatolykislov:
…tors hashing filter (APP-5, APP-6)


src/main/kotlin/app/extractors/JavaExtractor.kt
```diff
@@ -0,0 +1,57 @@
...
+
+package app.extractors
+
+import app.model.DiffContent
+import app.model.Stats
+
+class JavaExtractor : ExtractorInterface {
+    val NAME = "Java"
+
+    val KEYWORDS = listOf("abstract", "continue", "for", "new", "switch",
```
sergey-surkov: Ummm... why are you doing this? We did not make a product decision to parse out keywords. We discussed it as one of several ideas, and while it'a neat idea, we need to have Ryan's buy-in for this.

src/main/kotlin/app/model/DiffContent.kt
```diff
@@ -0,0 +1,16 @@
...
+package app.model
+
+import java.nio.file.Path
+import java.nio.file.Paths
+
+/**
+ * Per file diff from commit.
+ */
+data class DiffContent(
+        var path: Path = Paths.get(""),
```
sergey-surkov: I think it's fine for now, but in future, we will need to track more info so that extractors could make better decisions. For instance, antlr will need an entire file to parse out, so the extractor will need line numbers that we were added or deleted, not the lines themselves.

src/main/kotlin/app/utils/FileHelper.kt
```diff
@@ -0,0 +1,14 @@
...
+// Author: Anatoly Kislov (anatoly@sourcerer.io)
+
+package app.utils
+
+import java.nio.file.Path
+
+object FileHelper {
+    fun getFileExtension(path: Path): String {
+        val fileName = path.fileName.toString()
+        return fileName.substringAfterLast(
```
sergey-surkov: Hard to believe there isn't a standard java library that does that.

src/main/kotlin/app/utils/UiHelper.kt
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017 Sourcerer Inc. All Rights Reserved.
+// Author: Anatoly Kislov (anatoly@sourcerer.io)
+
+package app.utils
+
+object UiHelper {
+    fun confirm(message: String): Boolean {
```
sergey-surkov: It's usually a nice idea to require precisely 'y' or 'n', and if the input is something else, it makes sense to repeat the question. An alternative is to have a default (configurable through the call to UIHelper.confirm) that will capitalize the default choice e.g. "y/N"

src/main/kotlin/app/Main.kt
```diff
@@ -103,8 +105,8 @@ fun doRemove(commandRemove: CommandRemove) {
 
 fun doSetup() {
     if (!Configurator.isFirstLaunch()) {
-        println("Are you sure that you want to setup Sourcerer again? [y/n]")
-        if ((readLine() ?: "").toLowerCase() == "y") {
+        if(UiHelper.confirm("Are you sure that you want to setup Sourcerer "
```
anatolykislov: Ok.

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -90,91 +85,158 @@ class RepoHasher(val localRepo: LocalRepo) {
...
-            }
-            if (removedCommits.contains(it)) {
-                removedCommits.remove(it)
-            } else {
-                addedCommits.add(it)
+        return Configurator.getRepos()
+            .find { it.rehash == repo.rehash } != null
+    }
+
+    private fun findFirstOverlapCommit(): Commit? {
```
anatolykislov: Ok.

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -90,91 +85,158 @@ class RepoHasher(val localRepo: LocalRepo) {
...
+            .blockingFirst(null)
+    }
+
+    private fun hashAndSendCommits() {
+        val lastKnownCommit = repo.commits.lastOrNull()
+        val knownCommits = repo.commits.toHashSet()
+        getObservableCommits()
+            .pairWithNext()  // Pair commits to get diff.
+            .takeWhile { (new, _) ->  // Hash until last requested commit.
+                new.rehash != lastKnownCommit?.rehash }
```
anatolykislov: It's backward. We take after last known commit but we start from latest commit in master and go further. "Requested" is mistake, should be "known".

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -90,91 +85,158 @@ class RepoHasher(val localRepo: LocalRepo) {
...
+            .filter { (new, _) -> knownCommits.isEmpty()  // Don't hash known.
```
anatolykislov: Convention of wrapping lines before non-assignment operators are used. Moving || will break consistency.

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -90,91 +85,158 @@ class RepoHasher(val localRepo: LocalRepo) {
...
+                || !knownCommits.contains(new) }
+            .filter { (new, _) -> emailFilter(new) }  // Email filtering.
+            .map { (new, old) ->  // Mapping and stats extracting.
```
anatolykislov: Ok.

src/main/kotlin/app/RepoHasher.kt
```diff
@@ -90,91 +85,158 @@ class RepoHasher(val localRepo: LocalRepo) {
...
+                val diffContents = getDiffContents(new, old)
+                Logger.debug("Commit: ${new.raw?.name ?: ""}: "
+                    + new.raw?.shortMessage)
+                Logger.debug("Diff: " + diffContents.joinToString("\n"))
```
anatolykislov: Probably yes. I'll leave printing of only first entry.

src/main/kotlin/app/extractors/JavaExtractor.kt
anatolykislov: It's more for testing purposes. It isn't accurate and it was easiest feature to implement for more complex stats than just language stats.

src/main/kotlin/app/model/DiffContent.kt
anatolykislov: Got it.

src/main/kotlin/app/utils/FileHelper.kt
anatolykislov: For me too. I could add apache commons io for this simple function, but it looks like overkill.

src/main/kotlin/app/utils/UiHelper.kt
anatolykislov: Got it.

src/main/kotlin/app/extractors/JavaExtractor.kt
sergey-surkov: okay, sounds good.

src/main/kotlin/app/RepoHasher.kt
sergey-surkov: okay, i guess...

src/main/kotlin/app/utils/UiHelper.kt
```diff
@@ -4,8 +4,15 @@
 package app.utils
 
 object UiHelper {
-    fun confirm(message: String): Boolean {
-        println("$message [y/n]")
-        return (readLine() ?: "").toLowerCase() == "y"
+    fun confirm(message: String, default: Boolean): Boolean {
```
sergey-surkov: make it explicit: defaultIsYes

src/main/kotlin/app/utils/UiHelper.kt
```diff
@@ -4,8 +4,15 @@
...
+        val yes = "y"
```
sergey-surkov: This could be simplified to:
```
val yes = "Y" if defaultIsYes else "y";
val no = "N" if !defaultIsYes else "n";
println("$message [$yes/$no]);
```

PR #13, opened 2017-08-13
prototype code line age

asurkov:


src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+            val file = treeWalk.getPathString();
+            //println("found: ${file}");
+            val blameGen = BlameGenerator(repo, file)
+            blameGen.push(null, head)
+            val blame = blameGen.computeBlameResult()
+            val lines = blame.getResultContents()
+            var index = 0;
+            while (index < lines.size()) {
+                val blameCommit = blame.getSourceCommit(index)
+                val timeDiff = (head.getCommitTime() - blameCommit.getCommitTime())
```
anatolykislov: 80 column limit.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+    private fun computeDeletedLines(repo: FileRepository,
+                                    revWalk: RevWalk, head: RevCommit) {
+        revWalk.markStart(head);
+
+        var childCommit = head;
+        for (commit in revWalk) {
+            //println("\nparent $commit")
+            //println("this $childCommit")
+
+            val df = DiffFormatter(DisabledOutputStream.INSTANCE)
```
anatolykislov: Could be created outside of for loop.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+            df.setRepository(repo)
+            df.setDetectRenames(true)
+            val diffs = df.scan(commit, childCommit);
+            //println("filesChanged: ${diffs.size}")
+
+            for (diff in diffs) {
+                // todo: skip binary files.
+
+                // A new file, no deleted lines.
+                if (diff.changeType == DiffEntry.ChangeType.ADD) {
```
anatolykislov: We need only Delete and Modify. Skip Add, Remove, Copy.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+                            val authorEmail = author.getEmailAddress()
+
+                            list.add(CodeLine(blameCommit.getName(),
+                                     file, index,
+                                     authorName, authorEmail, timeDiff))
+                            index++
+                        }
+                    }
+
+                    delta += insCount - delCount
```
anatolykislov: With A begin and end we could identify deleted lines in old file, what's the point of using delta across different edits?

asurkov: I guess I misinterpreted begin/endA/B, something like a next chunk offsets are relative the original file, not against an applied previous chunk. Btw, do you have any good link where detailed explanations are given for these?

src/main/kotlin/app/CodeLongevity.kt
asurkov: I wish we had 2 spaces offsets rule, 4 spaces is so easy to break the limit

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+        val repo = FileRepository(repo_path)
+        val revWalk = RevWalk(repo);
+        val head = revWalk.parseCommit(repo.readOrigHead()); // master
+
+        computeExistingLines(repo, head)
+        computeDeletedLines(repo, revWalk, head)
+    }
+
+    fun test(email: String) {
+        var sum: Long = 0;
```
asurkov: I like this wrappy style, what does fold(0) stands for?

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Used to compute age of code lines in the repo.
+ */
+class CodeLongevity(repo_path: String) {
+    var repo_path: String = repo_path
+
+    var list: MutableList<CodeLine> = mutableListOf()
+
+    init {
+        println("READYY>>>>>>>>>>");
```
asurkov: you probably won't believe but I already removed dozens of those :)

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: I think in this case JGit's docs are pretty self-explanatory: [Edit(JGit)](http://download.eclipse.org/jgit/docs/jgit-2.0.0.201206130900-r/apidocs/org/eclipse/jgit/diff/Edit.html). A is for old content that was deleted, B is for new that was added. Begin and end is index of line (zero based). If begin equals end then no lines deleted from a or added to b.

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: Initial value.

src/main/kotlin/app/CodeLongevity.kt
asurkov: This all is good for a single chunk change case, it's not that clear when multiple chunks are involved, is it?

anatolykislov: Just indices, no offset.

asurkov: Ok. It's a side talk, but anyway.

I guess they can do at least two different types of indices to describe a diff: indices relative the original file and incremental indices (indices relative a result file after a previous chunk is applied).

Anyway, I'm happy to hear that you find with jgit documentation descriptive and self-explanatory :) I believe though they would benefit by providing couple examples to avoid possible misreadings.

anatolykislov: I mean it's explanatory in case of Edit not overall :)
Rüdiger Herrmann examples help a lot: [JGit diff](http://www.codeaffine.com/2016/06/16/jgit-diff/), [SO answers](https://stackoverflow.com/search?q=user:2986905+[jgit]).

asurkov: Still rumbling :) Yeah, I saw Herrmann's article, it's good and definitely better than nothing, but still. It's not even clear how they do a replace thing in case of beginA != beginB (if that's even possible scenario), because deletion and insertion are not commutative operations. And who on the earth suggested those awkward names beginA/beginB for deletions and insertions!

Jgit indeed feels with me as Gecko's XPCOM a decade ago.

anatolykislov: It is possible due to different number of lines added and deleted in previous partition edition of current file. JGit dealing with offsets instead of you. So you have Edit per partition edition of file. Begin/end a and begin/end b is absolute index of line for old and new file respectively.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+import org.eclipse.jgit.util.io.DisabledOutputStream
+
+import java.util.Date
+
+/**
+ * Used to store and identify a code line information in a repo.
+ */
+class CodeLine(val rev: String, val file: String, val line: Int,
+               val authorName: String, val authorEmail: String,
+               val age: Int) {
```
sergey-surkov: When integrating, please make sure that you use Anatoly's classes. (And feel free to adapt them). We only need one class for line information.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+import org.eclipse.jgit.treewalk.filter.*
+import org.eclipse.jgit.treewalk.TreeWalk
+import org.eclipse.jgit.util.io.DisabledOutputStream
+
+import java.util.Date
+
+/**
+ * Used to store and identify a code line information in a repo.
+ */
+class CodeLine(val rev: String, val file: String, val line: Int,
```
sergey-surkov: Please don't abbreviate. s/rev/revision/ (assuming it's a revision).

src/main/kotlin/app/CodeLongevity.kt
sergey-surkov: Obvs, this sort of statements should not be in the prod code.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    var list: MutableList<CodeLine> = mutableListOf()
+
+    init {
+        println("READYY>>>>>>>>>>");
+        println("Repo: $repo_path")
+
+        val repo = FileRepository(repo_path)
+        val revWalk = RevWalk(repo);
+        val head = revWalk.parseCommit(repo.readOrigHead()); // master
```
sergey-surkov: At least two spaces before the double slash in an inline comment. Also format:
```
);  // Master.
```

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: No, they are absolute indexes for old and new file.

src/main/kotlin/app/CodeLongevity.kt
sergey-surkov: I wonder if filter() would be way slower than a for loop. Don't know, but would be a cool test.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+    fun test(email: String) {
+        var sum: Long = 0;
+        for (line in list) {
+            if (line.authorEmail != email) {
+                continue
+            }
+            sum += line.age
+        }
+        if (list.size > 0) {
+            println("avg code line age for <$email> is ${sum / list.size} seconds")
```
sergey-surkov: When integrating, make sure things like this are under debugging mode logging.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Computes age of all existing lines in all files of the repo.
+     */
+    private fun computeExistingLines(repo: FileRepository, head: RevCommit) {
+        val treeWalk = TreeWalk(repo);
+        treeWalk.setRecursive(true);
+        treeWalk.addTree(head.getTree());
+        while (treeWalk.next()) {
+            // todo: skip binary files.
```
sergey-surkov: ```// TODO(alex): Skip binary files.```

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+
+            val file = treeWalk.getPathString();
+            //println("found: ${file}");
```
sergey-surkov: Please make sure commented code does not make it to master.

src/main/kotlin/app/CodeLongevity.kt
asurkov: This is for sure? because Herrman says "This list can be interpreted as the modifications that need to be applied to the old content in order to transform it into the new content", which seems to be a contradiction, for example, in case of two deletions.

anatolykislov: Sure. This sentence doesn't cover any indexes rules.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,173 @@
...
+                                  authorName, authorEmail, timeDiff))
+                index++
+            }
+        }
+    }
+
+    /**
+     * Computes age of all deleted lines in the repo.
+     */
+    private fun computeDeletedLines(repo: FileRepository,
```
sergey-surkov: Here is what worries me in this approach: (i) it potentially traverses repository more than once, (ii) it makes a counter-intuitive separation of lines that are present in head vs lines that were at some point deleted, (iii) it is not trivial to make it incremental.

Why not something cruder but possibly cleaner, e.g.:
```
- Maintain a hashmap: path -> list_of_lines, where each item in the list of lines is
   a couple (author, time)
- Iterate from the initial commit to head in the master (direct chronological order):
    - Iterate over all blobs in the commit:
        - if blob path is not in hash_map:
            - add a new entry: blob_path -> (commit_author, commit_date) *  N times,
               where N is the number of lines in the blob.
        - else:
            - for all deleted lines, emit (author, commit_time - time)
            - for all added lines, insert (commit_author, commit_time) into appropriate
               location in the list
```
Then we can save the arrays emitted during this operation, and simply restart the process once new commits appear on top of master.

We could also decide what to do with lines that were not overwritten yet. Those potentially have been living for a long time, so it would be unfair to discard them, so we can do something like (now - time) * 2, to give everybody some benefit of doubt and to recognize their lines are still alive in the latest and greatest version of the repo.

src/main/kotlin/app/CodeLongevity.kt
asurkov: Ok, if startA/endA are deletion indices against the old file, then that sentence appears to be a false one. Btw, how do you know what indices rules are?

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: No class for line of code yet.

src/main/kotlin/app/CodeLongevity.kt
asurkov: Well, I was asking for a feedback, not for a review for sure. Feel sorry for not making it clear from the start, it would save your time.

src/main/kotlin/app/CodeLongevity.kt
asurkov: yeah, that I wanted to revisit the alg by supporting the incremental updates (your concern #3) as a next iteration, it also supposed to get rid of blame usage (your concern #1).

src/main/kotlin/app/CodeLongevity.kt
asurkov: I'm good to have longer names, however this abbreviation goes close with jgit naming: they use RevCommit for example. Also, I use 'repo' not 'repository'. Are you sure about this change?

sergey-surkov: repo is a common slang. rev is not. Compare: https://en.wikipedia.org/wiki/Rev vs https://en.wikipedia.org/wiki/Repo_(disambiguation)

src/main/kotlin/app/CodeLongevity.kt
sergey-surkov: Right, right. I am just doing it in auto mode.

PR #16, opened 2017-08-15
feat: add MockApi for testing purposes (APP-21)

anatolykislov:
While testing field `api` of SourcererApi can be substituted with MockApi instance with specified mock of GET requests and values from POST and DELETE requests can be read after hashing complete.


src/main/kotlin/app/api/SourcererApi.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import app.model.Commit
+import app.model.Repo
+import app.model.User
+
+/**
+ * Sourcerer API.
+ */
+object SourcererApi : Api {
+    var api: Api = ServerApi()
```
sergey-surkov: Can you simply return a reference to api implementation if you want a singleton?

```
fun getInstance() { return api; }
```
and then you can delete all those methods that do precisely nothing.

anatolykislov: My concern is now we can use SourcererApi directly, e.g. `SourcererApi.postCommits(...)`, if we would use api through variable (getter could be set to variable directly) it would be `SourcererApi.api.postCommits(...)`.

Problems I see:
- It will change existing code where SourcererApi used
- Ugly signature
- Reference of api instance can be saved inside work class (e.g. RepoHasher) and later outside change of api will not affect anything.

sergey-surkov: Your SourcererAPI is essentially a factory. As far as a consumer is concerned, all you use it for is to produce an instance of API that varies depending on purpose (production/testing). To be fair, even a factor is not needed here. A traditional way to address the need for tests is to pass an API to its user constructor on initialization:

```
var api = new ServerAPI();
var user = new SomeAPIUser(api);
```

...and off you go. Very rarely in engineering there is a need for singletons, and it's usually for something pervasive like logging.

anatolykislov: Actually, SourcererAPI is such API user it is also pervasive because API used across app.

src/main/kotlin/app/ui/CloseState.kt
```diff
@@ -3,10 +3,13 @@
 
 package app.ui
```
sergey-surkov: why empty line?

anatolykislov: Same blocks as everywhere:
- Copyright
- Package
- Imports
- Class

PR #17, opened 2017-08-17
prototype code lifetime

asurkov:
here's a new version. not yet finished (I still need to sort out a rename thing), but I think it's ready for more feedback: code style and alg comments are welcome. I'm also not yet clear what exactly we want to land. I guess all those prints/test methods should go away or do we want to keep them until the code integrated and possibly covered by unit testing?

I didn't tried it on something heavy yet, so we might come with something more memory thrifty, for example, replace those line huge arrays on continues line chucks, if we find out it's a problem.


src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+ * Identifies a code line in a file in a revision.
+ */
+class RevCommitLine(val commit: RevCommit, val file: String, val line: Int)
+
+/**
+ * Represents a code line.
+ *
+ * TODO(Alex): the text arg is solely for testing proposes (remove it)
+ */
+class CodeLine(val from: RevCommitLine, val to: RevCommitLine, val text: String) {
```
sergey-surkov: Since our goal is to compute a distribution of line ages per author, I don't believe you need this class at all. Your algo only adds entries to it, and your test code iterates through all CodeLine instances, without regard to from/to, etc. You could simply keep track of author->array of line ages.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+
+                        var idx = insStart
+                        while (idx < insEnd) {
+                            val from = RevCommitLine(commit, newPath, idx)
+                            var to = newLines.get(idx)
+                            val cl = CodeLine(from, to, fileText.getString(idx))
+                            codeLines.add(cl)
+                            idx++
+                        }
+                        newLines.subList(insStart, insEnd).clear()
```
sergey-surkov: Okay, at this point, it's too many comments inline for me to understand what the code does :). I will wait for another iteration, and then comment more.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+                    }
+                }
+            }
+            commit = parentCommit
+        }
+        while (parentCommit != null && parentCommit != tail)
+
+        // If tail revision is specified then we have file code lines unclaimed,
```
sergey-surkov: So how do we treat lines that are still alive in the head revision?

src/main/kotlin/app/CodeLongevity.kt
asurkov: You have a bunch of related comments: this one, about oldId/newId one, probably the one about the alg is more complicated than it should be and what I'm doing about lines that are not yet deleted and couple more; I'll try to answer them here.

Existing lines are reported too, their age is the head's date minus date when they were inserted. When you got a living line and stored its age somewhere, then you will have to update its age on next iteration. That's why I have oldId and newId: newId in the head from previous iteration matches the oldId in the tail in a next iteration.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
+// Copyright 2017 Sourcerer Inc. All Rights Reserved.
```
asurkov: np, I'll file a follow up email on it

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+class RevCommitLine(val commit: RevCommit, val file: String, val line: Int)
+
+/**
+ * Represents a code line.
+ *
+ * TODO(Alex): the text arg is solely for testing proposes (remove it)
+ */
+class CodeLine(val from: RevCommitLine, val to: RevCommitLine, val text: String) {
+
+    // XXX(Alex): oldId and newId may be computed as a hash built from commit,
```
asurkov: np

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+        if (tailRev != "") RevWalk(repo).parseCommit(repo.resolve(tailRev))
+        else null
+
+    var codeLines: MutableList<CodeLine> = mutableListOf()
+
+    init {
+        compute()
+    }
+
+    fun test(email: String) {
```
asurkov: it's not a testing suit :) check out the pull request description, I left a more generic question about it

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+        // Build a map of file names and their code lines.
+        while (treeWalk.next()) {
+            // TODO(alex): skip binary files.
+            val fileName = treeWalk.getPathString()
+            val fileLoader = repo.open(treeWalk.getObjectId(0))
+            val fileText = RawText(fileLoader.getBytes())
+            val linesNum = fileText.size()
+            var lines = ArrayList<RevCommitLine>(linesNum)
+            var idx = 0
+            while (idx < fileText.size()) {
```
asurkov: as far as I know - for (variable in collection), they probably have a fancy way to create an array for indices, not sure though if they unwrap into same effective code. I'll leave that on @anatolykislov's plate if you don't mind.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+        df.setRepository(repo)
+        df.setDetectRenames(true)
+
+        val revWalk = RevWalk(repo)
+        revWalk.markStart(head)
+
+        var commit: RevCommit? = head
+        var parentCommit: RevCommit? = revWalk.next()  // proceed to head
+        do {
+            parentCommit = revWalk.next()
```
asurkov: this one and the one above: when you do markStart then you didn't moved the iterator yet. So I call the method twice. Ideas how to improve a comment are welcome :)

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+                }
+
+                // File was renamed, tweak the files map.
+                if (diff.changeType == DiffEntry.ChangeType.RENAME) {
+                    println("old path: ${diff.getOldPath()}")
+                    println("new path: ${diff.getNewPath()}")
+                    files.set(diff.getOldPath(),
+                              files.remove(diff.getNewPath())!!)
+                    println("oopsy")
+                    //continue
```
asurkov: check out the pull request description :) the renaming stuff is not yet implemented

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+                }
+
+                // Get a edit list and traverse it backwards to avoid indices
+                // adjustment.
```
asurkov: I have a bunch of ins/del chunks with indices relative original/new file, so I have to traverse the chunks array from back to use ins/del indices unchanged to ins/del to/from the lines array. Otherwise ins/del indices won't much indices of the lines array.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+                val editList = df.toFileHeader(diff).toEditList().asReversed()
+                for (edit in editList) {
+                    val delStart = edit.getBeginA()
+                    val delEnd = edit.getEndA()
+                    val delCount = edit.getLengthA()
```
asurkov: It's same, they provide then method, I use it. I'm ok with either way and can change it if find it more readable.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+                    if (delCount > 0) {
+                        val oldPath = diff.getOldPath()
+                        var tmpLines = ArrayList<RevCommitLine>(delCount)
+                        var idx = delStart;
+                        while (idx < delEnd) {
+                            tmpLines.add(RevCommitLine(commit, oldPath, idx));
+                            idx++
+                        }
+
+                        files.get(oldPath)!!.addAll(delCount, tmpLines)
```
asurkov: yeah, a typo :) thanks for the catch

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,231 @@
...
+    private fun compute() {
+        val treeWalk = TreeWalk(repo)
+        treeWalk.setRecursive(true)
+        treeWalk.addTree(head!!.getTree())
+
+        val files: MutableMap<String, ArrayList<RevCommitLine>> = mutableMapOf()
+
+        // Build a map of file names and their code lines.
+        while (treeWalk.next()) {
+            // TODO(alex): skip binary files.
```
asurkov: you mean the patch cannot be landed at master before it's addressed?

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,266 @@
...
+    // TODO(alex): oldId and newId may be computed as a hash built from commit,
+    // file name and line number, if we are going to send the data outside a
+    // local machine.
+
+    /**
+     * Id of the code line in a revision when the line was added. Used to
+     * identify a line and update its lifetime computed at the previous
+     * iteration.
+     */
+    val oldId: String = ""
```
asurkov: I think the server should keep (hash, age) pairs, which are used for age calculation for sure and to update age of living lines on a subsequent run. The server may but doesn't have to keep the records of deleted lines.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,266 @@
...
+        println("  last known as '${to.file}:${to.line}' in $tc $td")
+    }
+}
+
+/**
+ * Used to compute age of code lines in the repo.
+ */
+class CodeLongevity(repoPath: String, tailRev: String) {
+    val repo = FileRepository(repoPath)
+    val head: RevCommit = RevWalk(repo).parseCommit(repo.resolve("HEAD"));
```
asurkov: do you have on top of your head how to get a master branch? getOriginHead apparently doesn't work for local repos.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,266 @@
...
+            if (parentCommit != null) {
+                println("parent commit: ${parentCommit.getName()}; '${parentCommit.getShortMessage()}'")
+            }
+            else {
+                println("parent commit: null")
+            }
+
+            // A step back in commits history. Update the files map according
+            // to the diff.
+            val diffs = df.scan(parentCommit, commit)
```
asurkov: yes, that's the way to obtain a first commit changes / could you give an example?

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,266 @@
...
+                }
+
+                // TODO(alex): does it happen in the wilds?
+                if (diff.changeType == DiffEntry.ChangeType.COPY) {
+                    continue
+                }
+
+                // File was deleted, put its lines into the files map.
+                if (diff.changeType == DiffEntry.ChangeType.DELETE) {
+                    var lines = getFileLines(oldId, oldPath, commit)!!
```
asurkov: That shouldn't happen because we filter out binary files above so I use !! to terminate the program since something bad and unpredictable happening. Is NPE more suitable for this kind of things?

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: `repo.resolve("refs/heads/master")`

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Returns lines array of a text file of the given revision.
+     */
+    private fun getFileLines(fileId: ObjectId, filePath: String,
+                             commit: RevCommit) : ArrayList<RevCommitLine>? {
+        val fileLoader = repo.open(fileId)
+        if (RawText.isBinary(fileLoader.openStream())) {
+            return null
```
asurkov: It feels more complicated with me, null is sort of convenient way to report about error. I'm good with empty array as a result, however I would say the code will be less efficient. I don't think though it will be ever a bottleneck. So if you really prefer your way, then I can do it, otherwise I would keep it as is.

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: What's the point of storing them on the server? We could use local storage.

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: Ok.
```
var commit: RevCommit? = revWalk.next()
while (commit != null && commiy != tail) {
    var parentCommit: RevCommit? = revWalk.next()
    ...
    commit = parentCommit
}
```

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: I just want to say that ideology of a Kotlin is to get rid of a null references so I don't think that using null where it can be empty list is a good practice at least stdlib don't do it.

> Kotlin's type system is aimed at eliminating the danger of null references from code, also known as the The Billion Dollar Mistake. [Null Safety](https://kotlinlang.org/docs/reference/null-safety.htmll)

anatolykislov: And you double checking for binary format, in a mean of function reusability I think. If you want to do so isn't it better to throw exception `IllegalStateException("Binary file format")` it will terminate execution, it's self-explanatory and doesn't affect other code.

src/main/kotlin/app/CodeLongevity.kt
sergey-surkov: yeah, we should avoid sending anything to server except for stuff we can't do without. we don't need some blogger saying to everyone that sourcerer uploads per-line information of your proprietary code. if it can be done locally, let's do it locally.

asurkov: I'm not sure I see how commit info is different from line info when uploaded from a local machine in terms of blogger's speculations, but I guess lines can be stored and processed locally if we can set up a local DB for it.

src/main/kotlin/app/CodeLongevity.kt
asurkov: Kotlin indeed eliminated the danger of null references, and there's no way I could introduce it myself writing in kotlin :) the worst thing can happen is a program termination if I do !!

I can go with empty arrays, it's rather a matter of style, but it's ok.
I like the self explaining exceptions too, will use it

asurkov: On a second glance, having an empty array leaves us no way to distinguish empty files from binary files, so it'd be weird to throw a binary file format exception for empty files.
Also, I like self-describing exceptions, but I'd rather left a comment in the code in this case, because we have to add code (a check and exception throwing) for something that doesn't happen. There's no much sense in this in a production code.

src/main/kotlin/app/CodeLongevity.kt
sergey-surkov: Please no line info uploads to server. Local DB is fine assuming it does not introduce new dependencies to the app. A plain text file is probably the best.

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: Using exceptions inside function no empty array will be produced in case of binary files because function will be interrupted.

anatolykislov:
```
private fun getFileLines(fileId: ObjectId, filePath: String,
                             commit: RevCommit) : ArrayList<RevCommitLine> {
    val fileLoader = repo.open(fileId)
    if (RawText.isBinary(fileLoader.openStream())) {
        throw IllegalStateException("Binary file format")
    }
...
```
and
`var lines = getFileLines(oldId, oldPath, commit)`

src/main/kotlin/app/CodeLongevity.kt
asurkov: It's ok, however it's out of scope of this pull request, which was about alg to compute lines age. So, what's left before we can call it done? I assume, revert Main.kt changes, remove testing func, anything else? Or do you want to extend the scope to support lines age storage?

src/main/kotlin/app/CodeLongevity.kt
asurkov: It doesn't work because this function is used in two places, and one of them shouldn't fail.

anatolykislov: Why do you mean by shouldn't fail?

asurkov: I meant no exception. At first we collect the lines, and all binary files are ignored (no exception), then we reuse the function to collect the lines for deleted files, and no binary files is expected here, because we filtered them above, so if we run into a binary file at this point, then threre's something wrong with code above, that's why !! or exceptions.

I could inline the body of the helper function, if you like that more, at least that would address your concerns about null and !!. It might not look less nicer, but I can live with that.

anatolykislov: Basically what you do:
```
fun getFileLines(fileId, filePath, commit) : ArrayList? {
    if (RawText.isBinary(fileLoader.openStream())) {
        return null
    }
}

val lines = getFileLines(fileId, filePath, commit)
if (lines == null) {  // !!
    throw NullPointerException()
}
```

What you can do:
```
fun getFileLines(fileId, filePath, commit) : ArrayList? {
    if (RawText.isBinary(fileLoader.openStream())) {
        throw IllegalStateException()
    }
}

val lines = getFileLines(fileId, filePath, commit)  // Lines never null.
```

In that way when calling helper function there is no need to deal with nulls and exceptions anymore.

sergey-surkov: Okay, after carefully reading this debate, I'd like to break the tie. I agree with @anatolykislov for the following reasons: (i) it's better to have a descriptive exception, and (ii) any function needs to be written defensively and not assume the caller did a good job.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,267 @@
...
+                    codeLines.add(cl)
+                }
+            }
+        }
+    }
+
+    /**
+     * Returns lines array of a text file of the given revision.
+     */
+    private fun getFileLines(fileId: ObjectId, filePath: String,
```
sergey-surkov: Could you rename this? Usually 'get' accompanies functions that are 'light' in computing. This does a lot, so maybe readFileLines? or collectFileLines?

src/main/kotlin/app/CodeLongevity.kt
asurkov: just in case, it's not a wanted change, right? :)

src/main/kotlin/app/CodeLongevity.kt
asurkov: If I'm told twice that I'm wrong that I has to be wrong, let me defend myself though :). 

I'm not sure how I break (ii) since Kotlin takes care of it on a compiling stage, and not sure why we have to have an exception in the first place (i). I think the concept of exceptions is not quite suitable for this case: exceptions is a good way to say something wrong happens, mostly to react to unexpected (bad) input. If you need to catch a problem in your own code, then assertions concept is used. The assertions are seen in debug build only and totally ignored in release builds, and thus they have zero impact (cost) for release builds. Assertions unlike exceptions are used for something that should *never* happen, to debug your app. I believe that this case suits better for assertions than exceptions, regardless this concept is present in Kotlin or not.

So getFileLines is called twice: a) when the files map is created and b) when deleted files are processed. I'd say assertions is more appropriate for the latter case, and no exceptions is for the first case, otherwise we end up with try/catch constructions, which is somewhat ugly, because binary files is a normal thing for a common repo. You don't suggest to use exceptions mechanism as a return value, like hey here's is a binary file, and thus no lines for you, right?

Anyway, let me unwrap body of this function right into code, those few lines are not worth such long debates and probably don't even deserve its own function, at least it will be less ambiguous and not so concernful.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,267 @@
...
+  
+        val df = DiffFormatter(DisabledOutputStream.INSTANCE)
+        df.setRepository(repo)
+        df.setDetectRenames(true)
+
+        val revWalk = RevWalk(repo)
+        revWalk.markStart(head)
+
+        var commit: RevCommit? = revWalk.next()  // move the walker to the head
+        while (commit != null && commit != tail) {
```
asurkov: if tail it not empty and not a tail (just a random revision), then we have to crash, no?

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,267 @@
...
+            var parentCommit: RevCommit? = revWalk.next()
+
+            println("commit: ${commit!!.getName()}; '${commit.getShortMessage()}'")
```
asurkov: just a quick hint for me: what goes into logs?

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,267 @@
...
+            for (diff in diffs) {
+                val oldPath = diff.getOldPath()
+                val oldId = diff.getOldId().toObjectId()
+                val newPath = diff.getNewPath()
+                val newId = diff.getNewId().toObjectId()
+                println("old: '$oldPath', new: '$newPath'")
+
+                // Skip binary files.
+                var fileId =
+                    if (newPath != DiffEntry.DEV_NULL) newId else oldId
```
asurkov: how do you know that?! it's exactly 80 chars

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,267 @@
...
+                if (RawText.isBinary(repo.open(fileId).openStream())) {
```
asurkov: I've got impression they do a good job on it, they seem to analyze the first few bits only, but let's see how it goes, the binary files might be not that bottleneck we'll struggle with.

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -0,0 +1,267 @@
...
+                    var insEnd = edit.getEndB()
+                    val insCount = edit.getLengthB()
+                    println("del ($delStart, $delEnd), ins ($insStart, $insEnd)")
+
+                    // Deletion case. Chase down the deleted lines through the
+                    // history.
+                    if (delCount > 0) {
+                        var tmpLines = ArrayList<RevCommitLine>(delCount)
+                        var idx = delStart
+                        while (idx < delEnd) {
```
asurkov: where do you find those!

src/main/kotlin/app/CodeLongevity.kt
anatolykislov: @asurkov yes, you right, keep it then

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -172,11 +176,16 @@ class CodeLongevity(repoPath: String, tailRev: String) {
 
                 // File was deleted, put its lines into the files map.
                 if (diff.changeType == DiffEntry.ChangeType.DELETE) {
-                    // Text files only at this point, throw if not.
-                    var lines = getFileLines(oldId, oldPath, commit)!!
+                    val fileLoader = repo.open(oldId)
```
sergey-surkov: I actually liked it better in a separate function.

asurkov: on a positive side, there's not double binary check anymore

src/main/kotlin/app/CodeLongevity.kt
```diff
@@ -96,7 +96,7 @@ class CodeLongevity(repoPath: String, tailRev: String) {
             if (author.getEmailAddress() != email) {
                 continue
             }
-            line.printme()
+            println(line.toString())
```
anatolykislov: In prod logger should be used for debug output.

asurkov: I'll remove those on integration stage I think

PR #23, opened 2017-08-30
feat: add Fact, add sending longevity facts to server, rename Stats to CommitStats, rename Commits to CommitsGroup (APP-74)

anatolykislov:


src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -98,24 +103,51 @@ class CodeLongevity(private val localRepo: LocalRepo,
...
+            sums[email] = sums[email]!! + line.age
+            totals[email] = totals[email]!! + 1
         }
 
+        val repoAvg = if (repoTotal > 0) { repoSum / repoTotal } else 0
+        val stats = mutableListOf<Fact>()
+        stats.add(Fact(repo = serverRepo,
+                       key = CODE_LONGEVITY_REPO_KEY,
+                       value = repoAvg.toDouble()))
+        println("Repo average code line age is $repoAvg seconds, "
```
sergey-surkov: Shouldn't this be log? And why seconds? It should be days at least :)

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -98,24 +103,51 @@ class CodeLongevity(private val localRepo: LocalRepo,
...
-            println("Average code line age for <$email> is $avg seconds, "
-                  + "lines total: $total")
+            val total = totals[email] ?: 0
+            val avg = if (total > 0) { sums[email]!! / total } else 0
+            stats.add(Fact(repo = serverRepo,
+                           key = CODE_LONGEVITY_KEY,
+                           value = avg.toDouble(),
+                           author = Author(email = email)))
+            if (email == localRepo.author.email) {
+                println("Your average code line age is $avg seconds, "
```
sergey-surkov: log?

src/main/kotlin/app/model/Fact.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.security.InvalidParameterException
+
+/**
+ * Key-value statistics for fun facts.
+ */
+data class Fact(
+    var repo: Repo = Repo(),
+    var key: String = "",
+    var value: Double = 0.0,
+    var author: Author = Author()
```
sergey-surkov: I do think this should have an optional expiration date.

src/main/kotlin/app/hashers/CodeLongevity.kt
anatolykislov: Yep, you're right.

src/main/kotlin/app/hashers/CodeLongevity.kt
anatolykislov: Same.

src/main/kotlin/app/model/Fact.kt
anatolykislov: Better to use DB timestamps for expiration calculation.

sergey-surkov: I disagree. Timestamp could depend on what a fact is. And it's better to keep fact-specific knowledge to as few places as possible. The app knows best when a fact should expire.

anatolykislov: Could you give an example for facts with different expiration interval?

sergey-surkov: Some facts can be transient. E.g. "I just used C++11 for the first time". This should expire in a month or so.

sergey-surkov:
BTW, feel free to merge it. Expiration date on fun facts can always be added later. It's not a showstopper, that's why I approved it yesterday.

src/main/kotlin/app/model/Fact.kt
anatolykislov: Good example. Let's discuss it offline.

PR #24, opened 2017-09-04
feat: add all supported languages, add support for classifies (APP-78)

anatolykislov:
In this PR I changed the way how diff of commits pass to extractors. Because tech classifier will depend on recognized imports we need to send full content of files and specify diff ranges.
Add extracting of all supported languages via CommonExtractor. I think that we need specific extractor per each language according to some planning fun facts though.


yaronskaya:
What about html files where might be js code too?

anatolykislov:
Such case requires parsing of html. This functional will be introduced in next PRs.

PR #29, opened 2017-09-09
fix(longevity): 'to' commit may be computed wrongly for deleted lines…

asurkov:
do not add lines for deleted files since their lines will be added when a diff is processed (double addition)


anatolykislov:
Should this PR be closed?

asurkov:
no, why?

anatolykislov:
Because #30 contains the commit from this PR and new commit. Why did you split #29 and #30 ?

asurkov:
Right, it's on top of this branch. That's a thing I don't like about git, you cannot work on on top of other work and hide that fact from your reviewer and probably not confuse him :)

I follow 'one problem - one pull request' rule, it's much easier for tracking, reviewing and all.

So is my understanding correct that you approved both pull requests by approving #30?

anatolykislov:
Approved both.

PR #30, opened 2017-09-09
Longevity del2

asurkov:


src/test/kotlin/test/tests/hashers/CodeLongevityTest.kt
```diff
@@ -138,12 +138,92 @@ class CodeLongevityTest : Spek({
...
+    given("'test group #2'") {
+
+        val testRepoPath = "../testrepo2"
+        val testRepo = TestRepo(testRepoPath)
+        val fileName = "test1.txt"
+
+        // t2.1: initial insertion
+        val fileContent = listOf(
+          "test1",
+          "Addition to test1.txt in middle (Commit4)",
```
anatolykislov: Better to use more formal contents for tests. Too random now.

asurkov: that was exact copy of your repo's changeset that shown the problem, I don't mind though to change it to anything you like, what content you would see there?

anatolykislov: Yes, I know. But it was in local repo not in code. Something like:
- "Line 1",
- "Line 2",
- ...

PR #33, opened 2017-09-16
chore(longevity): use observable to iterate over lines (APP-100)

asurkov:


asurkov:
I do like shorter names as long as they stay descriptive, but if the convention is to prefix methods by 'get' or something, then it's fine with me.

anatolykislov:
[Soft convention](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names).

PR #35, opened 2017-09-20
Lib classifiers 

yaronskaya:


src/main/kotlin/app/extractors/PythonExtractor.kt
```diff
@@ -6,13 +6,25 @@ package app.extractors
...
+import org.jpmml.evaluator.ModelEvaluatorFactory
+import org.jpmml.evaluator.VoteDistribution
+import org.jpmml.model.PMMLUtil
+import java.io.File
 
 class PythonExtractor : ExtractorInterface {
     companion object {
         val LANGUAGE_NAME = "python"
         val FILE_EXTS = listOf("py", "py3")
+        val LIBRARIES_CLASSIFIER_PATH = "data/models/python_libs_pipeline.pmml"
```
anatolykislov: Add this file to Git.

src/main/kotlin/app/extractors/PythonExtractor.kt
```diff
@@ -6,13 +6,25 @@ package app.extractors
...
+        val LIBRARIES = listOf<String>() // TODO: add loading list of libraries
     }
 
+    val pmml = PMMLUtil.unmarshal(File(LIBRARIES_CLASSIFIER_PATH).inputStream())
+    val evaluator = ModelEvaluatorFactory.newInstance()
+                    .newModelEvaluator(pmml)
```
anatolykislov: Better to align on dot or just 4 spaces if it doesn't fit in a such way.

src/main/kotlin/app/extractors/PythonExtractor.kt
```diff
@@ -34,4 +46,19 @@ class PythonExtractor : ExtractorInterface {
 
         return libraries.toList()
     }
+
+    override fun getLineLibrary(line: String): String {
+        val arguments = LinkedHashMap<FieldName, FieldValue>()
+
+        for (inputField in evaluator.inputFields) {
```
anatolykislov: Fields are all known libraries?

src/main/kotlin/app/extractors/PythonExtractor.kt
```diff
@@ -34,4 +46,19 @@ class PythonExtractor : ExtractorInterface {
 
         return libraries.toList()
     }
+
+    override fun getLineLibrary(line: String): String {
```
anatolykislov: This function not used and it doesn't account detected imports. Did it make intentionally?

yaronskaya: It's a basic piece that returns the result of a classifier which we then should use in commit stat aggregating by intersection with detected imports.

anatolykislov: I think that it's better to select library according to detected imports inside this function because in such way we don't lose library that is actually imported but has smaller score than a library which isn't imported.

yaronskaya: I understand it, but my current working approach is when the classifier returns bunch of libraries based on the output probabilities(if some libraries have almost the same probability as the library with the highest prob we return them too). Anyway I'll think of how to improve the quality by introducing imported libraries. Thanks!

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -90,15 +104,51 @@ interface ExtractorInterface {
...
 
     fun tokenize(line: String): List<String> {
-        val regex =
-            Regex("""\s|,|;|\*|\n|\(|\)|\[|]|\{|}|\+|=|&|\$|!=|\.|>|<|#|@|:|\?""")
-        val tokens = regex.split(line)
-            .filter { it.isNotBlank() && !it.contains('"') && !it.contains('"')
-                && it != "-"}
+        val stringRegex = Regex("""(".+?"|'.+?')""")
+        val newLine = stringRegex.replace(line, "")
+        //TODO: multiline comment regex
```
sergey-surkov: TODO(lyaronskaya)

PR #41, opened 2017-09-28
feat: google analytics (APP-105)

anatolykislov:


src/main/kotlin/app/Analytics.kt
```diff
@@ -0,0 +1,95 @@
+package app
```
sergey-surkov: header

src/main/kotlin/app/Analytics.kt
```diff
@@ -0,0 +1,95 @@
...
+    private val TRACKING_ID = BuildConfig.GA_TRACKING_ID
+    private val DATA_SOURCE = "app"
+
+    private val HIT_PAGEVIEW = "pageview"
+    private val HIT_EXCEPTION = "exception"
+
+    private val fuelManager = FuelManager()
+
+    var uuid: String = ""  // Should be set on start of the app.
+    var username: String = ""  // Should be set on success authorization.
```
sergey-surkov: successful?

src/main/kotlin/app/Analytics.kt
```diff
@@ -0,0 +1,95 @@
...
+                                   "ds" to DATA_SOURCE,
+                                   "t" to HIT_PAGEVIEW,
+                                   "dp" to event)
+
+        try {
+            // Send event to GA with united params.
+            post(params + defaultParams.filter { !params.contains(it) }
+                + idParams).responseString()
+        } catch (e: Throwable) {
+            // Never fail from analytics.
```
sergey-surkov: Add some sort of log/console output here, at least.

src/main/kotlin/app/Analytics.kt
```diff
@@ -0,0 +1,95 @@
...
+        val defaultParams = listOf("v" to PROTOCOL_VERSION,
+                                   "tid" to TRACKING_ID,
+                                   "ds" to DATA_SOURCE,
+                                   "t" to HIT_PAGEVIEW,
+                                   "dp" to event)
+
+        try {
+            // Send event to GA with united params.
+            post(params + defaultParams.filter { !params.contains(it) }
+                + idParams).responseString()
```
sergey-surkov: weird line breakdown

src/main/kotlin/app/Analytics.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    init {
+        fuelManager.basePath = BASE_PATH
+    }
+
+    private fun post(params: List<Param>): Request {
+        return fuelManager.request(Method.POST, "/collect", params)
+    }
+
+    private fun trackEvent(event: String, params: List<Param> = listOf()) {
```
sergey-surkov: Can you add a comment that explains how exactly this is being logged? E.g. if it's virtual urls, can you explain how it will map to virtual urls? I also suggest all virtual urls start with /virtual prefix as a convention.

src/main/kotlin/app/Analytics.kt
```diff
@@ -0,0 +1,95 @@
...
+        trackEvent("/hashing/repo/success")
+    }
+
+    fun trackHashingSuccess() {
+        trackEvent("/hashing/success")
+    }
+
+    fun trackError() {
+        // TODO(anatoly): Send message with exd, consider limit 150 bytes.
+        trackEvent("/error", listOf("t" to HIT_EXCEPTION))
```
sergey-surkov: there should be some sort sort of error code as an argument for trackError, e.g.

/virtual/app/error/out-of-disk-space

src/main/kotlin/app/Analytics.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    fun trackEnd() {
+        trackEvent("/end")
```
sergey-surkov: is this app-exit? :)

anatolykislov: Yeah.

src/main/kotlin/app/Analytics.kt
anatolykislov: Todo is about it.

src/main/kotlin/app/Analytics.kt
anatolykislov: Why?

src/main/kotlin/app/Analytics.kt
anatolykislov: Ok.

src/main/kotlin/app/Analytics.kt
anatolykislov: Yes.

src/main/kotlin/app/Analytics.kt
anatolykislov: Ok.

src/main/kotlin/app/Analytics.kt
```diff
@@ -61,35 +77,35 @@ object Analytics {
...
 
     fun trackHashingSuccess() {
-        trackEvent("/hashing/success")
+        trackEvent("hashing/success")
     }
 
     fun trackError() {
         // TODO(anatoly): Send message with exd, consider limit 150 bytes.
-        trackEvent("/error", listOf("t" to HIT_EXCEPTION))
+        trackEvent("error", listOf("t" to HIT_EXCEPTION))
```
sergey-surkov: Normally, you don't want to change your logged events because those that are logged will stay in GA forever. So any change to log format, like renaming events, will impact funnels and their historical availability. I will let this one slip this one time, but please follow up with detailed error codes asap.

anatolykislov: I think it's better to use special error tracking software, e.g. https://sentry.io, because error code is not enough.

anatolykislov: For GA error urls probably best solution is to use exception names, agree?

PR #43, opened 2017-10-03
feat(longevity): implement line age accumulation and update (APP-71)

asurkov:


src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -167,6 +210,57 @@ class CodeLongevity(private val serverRepo: Repo,
     }
 
     /**
+     * Scans the repo to extract code line ages.
+     */
+    fun scan() : CodeLineAges? {
+        // Load existing age data if any.
+        val istream = try { ObjectInputStream(FileInputStream(storagePath)) }
```
anatolykislov: iStream (naming conventions)

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -167,6 +210,57 @@ class CodeLongevity(private val serverRepo: Repo,
...
+        // Update ages.
+        getLinesObservable(storedHead).blockingSubscribe { line ->
+            Logger.debug("Scanning: ${line}")
+            if (line.to.isDeleted) {
+                var age = line.age;
+                if (ageData.lastingLines.contains(line.oldId)) {
+                    age += ageData.lastingLines.remove(line.oldId)!!.age
+                }
+                val aggrAge = ageData.aggrAges.getOrPut(line.email,
+                                                        { CodeLineAges.AggrAge() } )
```
anatolykislov: It's ok to use 4 space ident (and not aligning with function arguments) in case a line doesn't fit 80 column limit

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -152,8 +195,8 @@ class CodeLongevity(private val serverRepo: Repo,
               + "lines total: $repoTotal")
 
         for (email in emails) {
-            val total = totals[email] ?: 0
-            val avg = if (total > 0) { sums[email]!! / total } else 0
+            val aggrAge = aggrAges[email] ?: CodeLineAges.AggrAge()
+            val avg = if (aggrAge.count > 0) { aggrAge.sum / aggrAge.count } else 0
```
anatolykislov: Doesn't fit 80 column limit.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -103,44 +111,79 @@ class CodeLine(val repo: Repository,
...
-    fun update() {
-        val sums: MutableMap<String, Long> = emails.associate { Pair(it, 0L) }
-                                                   .toMutableMap()
-        val totals: MutableMap<String, Int> = emails.associate { Pair(it, 0) }
-                                                    .toMutableMap()
+    /**
+     * Update code line age statistics on the server.
+     */
+    fun updateStats(api: Api) {
+        var ages = scan()
```
anatolykislov: val ages = scan() ?: return

src/test/kotlin/test/tests/hashers/CodeLongevityTest.kt
```diff
@@ -353,4 +351,123 @@ class CodeLongevityTest : Spek({
...
+    given("'line storage #1'") {
+        val testRepoPath = "../CodeLongevity_ls1"
+        val testRepo = TestRepo(testRepoPath)
+        val testRehash = "rehash_ls1"
+        val fileName = "test1.txt"
+        var email = testRepo.userEmail
+        val emails = hashSetOf(email)
+
+        testRepo.createFile(fileName, listOf("line1", "line2"))
+        testRepo.commit(message = "inital commit",
```
anatolykislov: initial

src/test/kotlin/test/tests/hashers/CodeLongevityTest.kt
```diff
@@ -353,4 +351,123 @@ class CodeLongevityTest : Spek({
             testRepo.destroy()
         }
     }
+
+    given("'line storage #1'") {
+        val testRepoPath = "../CodeLongevity_ls1"
+        val testRepo = TestRepo(testRepoPath)
+        val testRehash = "rehash_ls1"
+        val fileName = "test1.txt"
+        var email = testRepo.userEmail
```
anatolykislov: val email

src/test/kotlin/test/tests/hashers/CodeLongevityTest.kt
```diff
@@ -12,13 +13,17 @@ import app.model.*
...
+import kotlin.test.assertTrue
 
 import org.eclipse.jgit.revwalk.RevCommit
 
 import org.jetbrains.spek.api.Spek
 import org.jetbrains.spek.api.dsl.given
 import org.jetbrains.spek.api.dsl.it
 
+import java.util.Calendar
+import java.util.Calendar.Builder
```
anatolykislov: Don't need this import.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -167,6 +210,57 @@ class CodeLongevity(private val serverRepo: Repo,
     }
 
     /**
+     * Scans the repo to extract code line ages.
+     */
+    fun scan() : CodeLineAges? {
+        // Load existing age data if any.
+        val istream = try { ObjectInputStream(FileInputStream(storagePath)) }
+            catch(e : Exception ) { null }
+        var storedHead = istream?.readObject() as RevCommit?
```
anatolykislov: val storedHead

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -167,6 +210,57 @@ class CodeLongevity(private val serverRepo: Repo,
...
+                if (ageData.lastingLines.contains(line.oldId)) {
+                    age += ageData.lastingLines.remove(line.oldId)!!.age
+                }
+                ageData.lastingLines.put(line.newId,
+                                         CodeLineAges.LineInfo(age, line.email))
+            }
+        }
+
+        // Store ages for subsequent runs.
+        val ostream = ObjectOutputStream(FileOutputStream(storagePath))
```
anatolykislov: oStream. Handle exceptions and create directories if not presented here.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -167,6 +210,57 @@ class CodeLongevity(private val serverRepo: Repo,
...
+            return null
+        }
+
+        val ageData = (istream?.readObject() ?: CodeLineAges()) as CodeLineAges
+
+        // Update ages.
+        getLinesObservable(storedHead).blockingSubscribe { line ->
+            Logger.debug("Scanning: ${line}")
+            if (line.to.isDeleted) {
+                var age = line.age;
```
anatolykislov: Excess semicolon.

src/test/kotlin/test/tests/hashers/CodeLongevityTest.kt
```diff
@@ -353,4 +351,123 @@ class CodeLongevityTest : Spek({
             testRepo.destroy()
         }
     }
+
+    given("'line storage #1'") {
```
anatolykislov: This and next test group fails with java.io.FileNotFoundException if data/storage directory not presented.

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: I'd rather add an empty 'storage' directory into the repo by putting .gitignore into the dir

anatolykislov: This directories should be created on user machine, how does it solve the problem?

asurkov: The files are currently located at 'data/storage' in the repo. As far as I understand, the problem you observed was 'storage' folder was absent. If I add one into the repo, then the problem is solved, correct?

anatolykislov: Not really. ML models located at src/main/resources/data/libraries/ now. What that means is that files will be packaged into JAR container. This files can't be changed during runtime. We will distribute only JAR container and all directories that used for runtime generated files should be created programmatically.

asurkov: ok, what my bet would be for a path to store the files?

anatolykislov: Now we only saving config file .sourcerer in user directory. I think its time to create .sourcerer/ directory instead and store all files there.

asurkov: so ./.sourcerer/data/storage or ./.sourcerer/longevity or just ./.sourcerer?

anatolykislov: ./.sourcerer/incremental/longevity may be?

asurkov: replace incremental to more generic data?

anatolykislov: Ok

PR #49, opened 2017-10-21
fix(longevity): StackOverflowError on serializing (APP-126)

asurkov:


src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -146,7 +147,9 @@ class CodeLongevity(private val serverRepo: Repo,
                     git: Git) {
     val repo: Repository = git.repository
     val head: RevCommit =
-        RevWalk(repo).parseCommit(repo.resolve(RepoHelper.MASTER_BRANCH))
+        try { RevWalk(repo).parseCommit(repo.resolve("HEAD")) }
```
anatolykislov: All hashers should be consistent in what branch they hash. We have agreement of hashing master branch, ref defined in RepoHelper.MASTER_BRANCH.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -146,7 +147,9 @@ class CodeLongevity(private val serverRepo: Repo,
...
+        catch(e: Exception) { throw Exception("No default branch") }
```
anatolykislov: No master branch. But existence of master branch checked before run of hasher, so this check is excess.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -253,7 +262,7 @@ class CodeLongevity(private val serverRepo: Repo,
             Files.createDirectories(Paths.get(storageDir))
         }
         val oStream = ObjectOutputStream(FileOutputStream(storagePath))
```
anatolykislov: This call should be encapsulated with try-catch, cuz can produce exceptions.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -215,15 +218,21 @@ class CodeLongevity(private val serverRepo: Repo,
...
+            Logger.debug("Stored repo head: storedHeadId")
+            storedHead = RevWalk(repo).parseCommit(repo.resolve(storedHeadId))
+            if (storedHead == head) {
+                return null
+            }
+            ageData = (iStream.readObject() ?: CodeLineAges()) as CodeLineAges
         }
-
-        val ageData = (iStream?.readObject() ?: CodeLineAges()) as CodeLineAges
+        catch(e: FileNotFoundException) { }
```
anatolykislov: ObjectInputStream and FileInputStream can possibly produce a lot of different exceptions, so e: Exception is more appropriate as it generic. We should fail hashing in case of problems with loading of data for incremental updates.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -215,15 +218,21 @@ class CodeLongevity(private val serverRepo: Repo,
...
         // Load existing age data if any.
-        val iStream = try { ObjectInputStream(FileInputStream(storagePath)) }
-            catch(e : Exception ) { null }
-        val storedHead = iStream?.readObject() as RevCommit?
-        if (storedHead == head) {
-            return null
+        try {
+            val iStream = ObjectInputStream(FileInputStream(storagePath))
+            val storedHeadId = iStream.readUTF()
+            Logger.debug("Stored repo head: storedHeadId")
```
anatolykislov: Missed $ for substitution.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -215,15 +218,21 @@ class CodeLongevity(private val serverRepo: Repo,
...
+
         // Load existing age data if any.
-        val iStream = try { ObjectInputStream(FileInputStream(storagePath)) }
-            catch(e : Exception ) { null }
-        val storedHead = iStream?.readObject() as RevCommit?
-        if (storedHead == head) {
-            return null
+        try {
+            val iStream = ObjectInputStream(FileInputStream(storagePath))
+            val storedHeadId = iStream.readUTF()
```
anatolykislov: Maybe add comment that this method read first line of file? Really strange naming of method they've made.

asurkov: Not sure whether it read the first line or not, it reads a string whatever format it has. I can add a comment like "Storage format: commit id and CodeLineAges structure following it" if you think it makes sense.

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: np, I'll revert the branch name change. However I think this requirement is too restrictive though and overcomplicates the things in general. I'll file a bug on it to have further discussion.

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: This done on purpose, if something bad or unexpected happens I want to know that. This bug was also found, because we don't hide exceptions. We may want to hide that on release builds, but that's definitely useful thing for testing and debugging.

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: same, what do you want to see under a catch block?

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: btw, I'd like to keep the try/catch thing if that was your point, because it's useful for testing outside the main infrastructure.

src/main/kotlin/app/hashers/CodeLongevity.kt
anatolykislov: Yeah, nice.

PR #50, opened 2017-10-21
feat: add sentry error reporting

anatolykislov:


src/main/kotlin/app/Logger.kt
```diff
@@ -3,69 +3,133 @@
...
 /**
  * Singleton class that logs events of different levels.
  */
 object Logger {
+    object Events {
+        val START = "start"
+        val AUTH = "auth"
+        val CONFIG_SETUP = "config/setup"
+        val CONFIG_CHANGED = "config/changed"
+        val HASHING_REPO_SUCCES = "hashing/repo/success"
```
yaronskaya: HASHING_REPO_SUCCESS

PR #52, opened 2017-10-24
feat: models compression, fix bugs in clfs (APP-121)

yaronskaya:


src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -34,11 +40,22 @@ interface ExtractorInterface {
             return libraries
         }
 
+        fun unpickleModel(path: String): PMML {
+            val stream = getResource(path)
+            val storage = CompressedInputStreamStorage(stream)
+            var pklObject = PickleUtil.unpickle(storage)
+            val pipeline = pklObject as Pipeline
+            pklObject = PMMLPipeline().setSteps(pipeline.getSteps())
+            val pmml = pklObject.encodePMML()
```
anatolykislov: Initial loading of js evaluator takes 3 minutes on my machine, it's ok for first launch, but would be better if we can optimize time for subsequent launches. What if we generate PMML files from pickle on the first launch and then use PMML files to instantiate evaluators, will it be faster?

yaronskaya: If one uses not all languages - it might lead to 200 mb as before.

anatolykislov: I think that it's ok to generate such files locally, the problem was with transmission of such big files via internet. Also, I'm suggesting to generate evaluators on demand.

yaronskaya: > I think that it's ok to generate such files locally, the problem was with transmission of such big files via internet.

I understand.

> Also, I'm suggesting to generate evaluators on demand.

I thought we already do that..

anatolykislov: And cache it to PMML files, see first comment

PR #54, opened 2017-10-31
Move models to google cloud (APP-128)

yaronskaya:


src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -54,17 +56,57 @@ interface ExtractorInterface {
             return pmml
         }
 
+        fun downloadModel(name: String, outputDir: String) {
```
anatolykislov: Should be private.

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -45,7 +47,7 @@ interface ExtractorInterface {
         }
 
         fun unpickleModel(path: String): PMML {
-            val stream = getResource(path)
+            val stream = FileInputStream(path)
```
anatolykislov: Process exceptions.

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -54,17 +56,57 @@ interface ExtractorInterface {
             return pmml
         }
 
+        fun downloadModel(name: String, outputDir: String) {
+            val url = "https://storage.googleapis.com/lib_models/pkl/$name.pkl.z"
```
anatolykislov: Link should be moved to build config.

build.gradle
```diff
@@ -55,6 +55,9 @@ buildConfig {
     buildConfigField 'String', 'SENTRY_DSN', 'https://0263d6473bd24a9ba40e25aa5fb0a242:c5451dc815074bff8ce3fb9f0851f2f5@sentry.io/233260'
     buildConfigField 'boolean', 'PRINT_STACK_TRACE', 'true'
 
+    //Models storage path
```
anatolykislov: // Models storage path.

build.gradle
```diff
@@ -55,6 +55,9 @@ buildConfig {
...
+    buildConfigField 'String', 'LIBRARY_MODELS_URL', 'https://storage.googleapis.com/lib_models/pkl/'
```
anatolykislov: Account version of models, e.g. `'https://storage.googleapis.com/sourcerer-models/v1/'`.
Also, we should use single google cloud account for servers and models.

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -54,22 +56,70 @@ interface ExtractorInterface {
...
+            }
+        }
+
+        private fun getPickleInputStream(pklPath: String): FileInputStream {
+            val pklStream = try {
+                FileInputStream(pklPath)
+            }
+            catch (e: Exception) {
+                Logger.error(e, "Failed to load $pklPath")
+                FileInputStream(pklPath)
```
anatolykislov: Actually, in case of problems with loading pickle files exception should be throw and handled in top classes, cuz in that case we can't extract data at all. Second call may help in some cases but not in all. Remove it.

build.gradle
makrusak: sourcerer-app/ml_models/v1 ?

yaronskaya: Moved.

PR #55, opened 2017-11-03
C libs classifier(APP-120)

yaronskaya:


src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -278,6 +278,10 @@ interface ExtractorInterface {
             return emptyList()
         }
 
+        if (languageLabel == "c" && languageLabel in selectedCategories) {
```
anatolykislov: What this block do?

yaronskaya: It is specific for C language.

anatolykislov: I see. But what it do? No lines libraries for c language?

yaronskaya: No, just for a line that has language label selected.

anatolykislov: Constant already defined: `if (languageLabel == CExtractor.LANGUAGE_NAME && ... )`
Also, add comments about percentile, please.

PR #57, opened 2017-11-04
Metrics

asurkov:
I don't know if we really need it, but I find these methods useful for bottlenecks analysis - just put a bunch of recordTime {} calls, which could be nested, and you will see timing of which block takes what time.

Note, the PR contains changes from another PR, just ignore them, you are interested in the last one.


src/main/kotlin/app/utils/Metrics.kt
```diff
@@ -0,0 +1,82 @@
...
+ * calls hierarchy.
+ */
+fun <T> recordTime(tag: String, block: () -> T) : T {
+    return Metrics.recordTime(tag, block)
+}
+
+/**
+ * Records a numeric metric of given tag.
+ */
+fun recordMetric(tag: String, metric: Long)
```
anatolykislov: What this function used for? Storing of numbers?

asurkov: right, just accumulating numeric values, for example, it's useful, when you're recording the timing of line insertions and removals, and you want to know the ratio of inserted and removed lines. It is a very basic feature for sure, but often helpful.

PR #59, opened 2017-11-07
APP-141: report Longevity's crashes to Logger

asurkov:


src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -330,7 +331,7 @@ class CodeLongevity(private val serverRepo: Repo,
             }
         }
 
-        getDiffsObservable(tail).blockingSubscribe { (commit, diffs) ->
+        getDiffsObservable(tail).subscribe( { (commit, diffs) ->
```
anatolykislov: Asynchronous subscribe. Is it working not in a test mode?

asurkov: You mean if it was called from Main, then yes, it seems so. The problem of blockingSubscribe is it collects all items before the items are delivered to a consumer, which is a bit slower than it could be.

asurkov: @anatolykislov if you are skeptical about subscribe part, then let's back it out and deal with it separately in different issue?

anatolykislov: Ok, let's do it

src/main/kotlin/app/hashers/RepoHasher.kt
```diff
@@ -80,7 +80,8 @@ class RepoHasher(private val localRepo: LocalRepo, private val api: Api,
 
             // TODO(anatoly): CodeLongevity hash from observable.
             try {
-                CodeLongevity(serverRepo, filteredEmails, git).updateStats(api)
+                CodeLongevity(serverRepo, filteredEmails, git, onError).
```
anatolykislov: It's better to move point to next line :)

PR #65, opened 2017-11-14
feat: sparse implementation of lib clfs and using protobufs (APP-122)

yaronskaya:


src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -47,18 +49,9 @@ interface ExtractorInterface {
...
-            val pipeline = pklObject as Pipeline
-            pklObject = PMMLPipeline().setSteps(pipeline.getSteps())
-            val pmml = pklObject.encodePMML()
-            return pmml
-        }
-
-        private fun downloadModel(name: String, outputDir: String) {
-            val url = BuildConfig.LIBRARY_MODELS_URL + "$name.pkl.z"
-            val outputPath = "$outputDir/$name.pkl.z"
+        private fun downloadModel(name: String, ext: String, outputDir: String) {
```
sergey-surkov: I don't think you need to split name and extension like that.

PR #67, opened 2017-11-17
wip: python imports

yaronskaya:
Added multiple imports for opencv in python and handling imports of ..._pb (proto packages).


src/main/kotlin/app/extractors/PythonExtractor.kt
```diff
@@ -35,7 +37,12 @@ class PythonExtractor : ExtractorInterface {
             }
         }
 
-        return imports.toList()
+        var libraries = imports.map { MULTI_IMPORT_TO_LIB.getOrDefault(it, it) }
+            .filter { !it.endsWith("pb")}.toMutableList()
```
anatolykislov: Why is that hardcoded? It's dirty

yaronskaya: What do you mean?

PR #147, opened 2017-12-13
feat: detect colleagues (APP-159)

asurkov:


src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -85,9 +91,21 @@ class CodeLine(val repo: Repository,
     /**
      * Email address of the line's author.
      */
-    val email : String
+    val authorEmail : String
```
sergey-surkov: I think we need to start thinking beyond email. It should be some sort of an email-derived id here. e.g. sha256(email). Eventually, we won't have emails in our database, so let's start preparing for it.

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -73,8 +73,14 @@ class CodeLine(val repo: Repository,
     /**
      * The code line's age in seconds.
      */
-    val age : Long
-        get() = (to.commit.commitTime - from.commit.commitTime).toLong()
+    var age : Long
+        get() {
+            if (_age == 0L) {
```
anatolykislov: Why using backing field instead of "field" keyword?

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -107,7 +125,68 @@ class CodeLine(val repo: Repository,
...
+        if (vicinity > line.age) {
+            dates.put(month, line.age)
+        }
+    }
+
+    fun updateStats() {
+        // TODO(alex): report the stats to the server
+        if (Logger.isDebug) {
+            map.forEach( { (email1, email2), dates ->
+                println("<$email1> <$email2>")
```
anatolykislov: Logger.debug

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -107,7 +125,68 @@ class CodeLine(val repo: Repository,
...
+                dates.forEach { month, vicinity ->
+                    println("  $month: $vicinity ms")
```
anatolykislov: Logger.debug

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -107,7 +125,68 @@ class CodeLine(val repo: Repository,
...
+    }
+
+    private var _age : Long = 0;
+}
+
+/**
+ * Detects colleagues and their 'work vicinity' from commits.
+ */
+class Colleagues {
+    private val map: HashMap<Pair<String, String>,
```
anatolykislov: In case of using such types could you please comment them?

src/main/kotlin/app/hashers/CodeLongevity.kt
```diff
@@ -107,7 +125,68 @@ class CodeLine(val repo: Repository,
...
+                    println("  $month: $vicinity ms")
+                }
+            } )
+        }
+    }
+
+    /**
+     * Return colleagues for the given email.
+     */
+    fun get(email: String) : List<Triple<String, String, Long>> {
```
anatolykislov: Comment types.

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: Not sure what changes are required? It rather looks like a separate task with me, possibly involving changes in other hashers.

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: not sure I have a good background on this and even familiar with this, what would be your recommendation?

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: ok, not big deal though as the code should go away

src/main/kotlin/app/hashers/CodeLongevity.kt
asurkov: per our chat: this patch doesn't send anything on the server yet, while hashing should happen before data is sent to the server, thus no changes are required for this PR.

Issue #130, opened 2017-12-08
Add rich Kotlin support


osilla:
This task is about parsing Kotlin tech stats

PR #148, opened 2017-12-13
feat: kotlin clf (#130) (APP-149)

yaronskaya:


src/main/kotlin/app/extractors/KotlinExtractor.kt
```diff
@@ -0,0 +1,58 @@
...
+        newLine = packageRegex.replace(newLine, "")
+        return super.tokenize(newLine)
+    }
+
+    override fun getLineLibraries(line: String,
+                                  fileLibraries: List<String>): List<String> {
+
+        return super.getLineLibraries(line, fileLibraries, evaluator, LANGUAGE_NAME)
+    }
+}
```
anatolykislov: New line

Issue #110, opened 2017-12-08
(Fun Fact) I mostly use Spaces overs Tabs for indentation


osilla:
I mostly use Spaces overs Tabs for indentation
or
I mostly use Tabs overs Spaces for indentation

PR #154, opened 2017-12-20
feat: indentation fact (#110)

yaronskaya:


src/main/kotlin/app/hashers/FactHasher.kt
```diff
@@ -121,6 +123,11 @@ class FactHasher(private val serverRepo: Repo = Repo(),
             fsVariableNaming[email]!![FactCodes.VARIABLE_NAMING_OTHER] +=
                 others
         }
+
+        fsIndentation[email]!![FactCodes.INDENTATION_SPACES] +=
+            lines.count { it.isNotBlank()  &&  it.startsWith(" ") && !it.contains("\t")}
```
anatolykislov: it.isNotBlank() unnecessary check

yaronskaya: Don't want to count lines without code

anatolykislov: Gotcha

Issue #192, opened 2018-02-05
Models versioning


yaronskaya:
Let the app to download updated models.

PR #193, opened 2018-02-05
feat: models versioning (#192)

yaronskaya:


src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -6,18 +6,23 @@ package app.extractors
 
 import app.BuildConfig
 import app.Logger
+import app.ModelsProtos.ModelsVersions
```
anatolykislov: Please, implement wrapper for ModelVersionGroup / ModelVersion protobuf messages like: [Classifier](https://github.com/sourcerer-io/sourcerer-app/blob/develop/src/main/kotlin/app/extractors/Classifier.kt) or [CommitGroup](https://github.com/sourcerer-io/sourcerer-app/blob/develop/src/main/kotlin/app/model/CommitGroup.kt). ExtractorInterface should not deal with encoding, it's already doing too much.

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -61,24 +66,95 @@ interface ExtractorInterface {
...
+                    }
+                }
+            }
+            catch (e: Exception) {
+                Logger.error(e, "Failed to download versions file")
+            }
+        }
+
+        @Throws(InvalidProtocolBufferException::class)
+        fun getModelsVersions(proto: ModelsVersions):
```
anatolykislov: Should be moved to protobuf messages wrapper.

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -61,24 +66,95 @@ interface ExtractorInterface {
...
         fun getLibraryClassifier(name: String): Classifier {
+            val fileName = name + pbExt
+
             if (classifiersCache.containsKey(name)) {
                 return classifiersCache[name]!!
             }
 
-            val fileName = name + pbExt
-            if (FileHelper.notExists(fileName, modelsDir)) {
+            val appVersionFileName = "app_versions.pb"
```
anatolykislov: Move constants to companion object.

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -61,24 +66,95 @@ interface ExtractorInterface {
...
+            val storageVersionFileName = "versions.pb"
+            if (FileHelper.notExists(appVersionFileName, modelsDir)) {
```
anatolykislov: Code for downloading and checking model version is better to move to different functions, e.g.:
```
fun getLibraryClassifier(name: String): Classifier {
  if (cache.containsKey(name)) { return cache[name] }
  if (FileHelper.notExists(name, modelsDir) || !isLastestModelVersion(name)) {
    downloadModel(name)  // This function should also update app model version file.
  }
  val model = readModel(name)
  cache[name] = model
  return model
}
```

src/main/kotlin/app/extractors/ExtractorInterface.kt
```diff
@@ -61,24 +66,95 @@ interface ExtractorInterface {
...
+                Logger.info { "Downloading " + storageVersionFileName }
+                downloadModelVersions(storageVersionFileName)
```
anatolykislov: Storage versions file should be downloaded on each app start to keep tracking for new model version. I'm suggesting to move initialization of storage version object to companion object.

Issue #165, opened 2018-01-08
Erroneous and missing technologies for user 


osilla:
comment from a recent user of ours.  note his issues on not listing SASS or CSS for him and that Ember and React are listed but he never used them.

https://sourcerer.io/adnanrahic

----

Great! Yeah, the process was very simple and straightforward. Only thing I'd see as an improvement in the CLI is autocompletion for directories. It was a bit of a pain to enter 15 repos by hand, one at a time, having to type out the absolute path for every one of them. :)Regarding the profile, it's spot on. But I can't seem to figure out why it's not showing any SASS. I use SASS for 3 of the largest repos I've contributed to. So it was a bit strange not to see any SASS or CSS at all.

And, it's a bit strange I see Ember and React among the technologies. I've never used any of those. :D |

anatolystansler:
Fixed

PR #196, opened 2018-02-05
fix: do not process js comments (#165)

yaronskaya:


src/main/kotlin/app/extractors/JavascriptExtractor.kt
```diff
@@ -24,12 +24,16 @@ class JavascriptExtractor : ExtractorInterface {
...
-            Regex("""\s+|,|;|:|\*|\n|\(|\)|\\[|]|\{|}|\+|=|\.|>|<|#|@|\$""")
+            Regex("""\s+|,|;|:|\*|\n|\(|\)|\[|]|\{|}|\+|=|\.|>|<|#|@|\$""")
+        val multilineCommentRegex = Regex("""/\*.+?\*/""")
         val twoOrMoreWordsRegex = Regex("""(".+?\s.+?"|'.+?\s.+?')""")
 
-        val line = fileContent.joinToString(separator = " ").toLowerCase()
-        val fileTokens = twoOrMoreWordsRegex.replace(line, "").split(splitRegex)
-
+        val commentRegex = Regex("""^([^\n]*//)[^\n]*""")
+        val line = fileContent.map { commentRegex.replace(it, "")}
```
anatolykislov: val line = fileContent.map { line -> commentRegex.replace(line, "") }

src/main/kotlin/app/extractors/JavascriptExtractor.kt
```diff
@@ -24,12 +24,16 @@ class JavascriptExtractor : ExtractorInterface {
 
     override fun extractImports(fileContent: List<String>): List<String> {
         val splitRegex =
```
anatolykislov: Let's move all regex variables to companion object, will increase performance.

PR #204, opened 2018-02-11
feat: team size calculation (#194)

yaronskaya:


yaronskaya:
I've added preprocessing of threegrams for names and emails. Will make it slightly faster.

The current threshold is a reasonable value to compare two strings that have same substrings. I think your data is based on knowledge of github about completely different emails that a user might have.

PR #210, opened 2018-02-17
feat: add sending of meta data for cloud hashing, refactoring

anatolykislov:


src/main/kotlin/app/ui/UpdateRepoState.kt
```diff
@@ -24,16 +24,15 @@ class UpdateRepoState constructor(private val context: Context,
     override fun doAction() {
         Logger.info { "Hashing started" }
 
+        // TODO(anatoly): Move code inside RepoHasher.
```
yaronskaya: to

PR #211, opened 2018-02-19
feat: team size and minor fixes

anatolykislov:
- Add correct team size calculation with filtering of similar emails and names
- Fixed default branch choosing (current order of branch search: remote's default branch, remote's master, local master, local HEAD)
- Recognize different case for emails
- Improved tech assigning for JS (Stopped processing of JS comments for extracting of tokens)

### DIFF
--- a/src/main/kotlin/app/hashers/FactHasher.kt
+++ b/src/main/kotlin/app/hashers/FactHasher.kt
@@ -8,7 +8,6 @@ import app.FactCodes
 import app.Logger
 import app.api.Api
 import app.extractors.Extractor
-import app.extractors.PythonExtractor
 import app.model.Author
 import app.model.Commit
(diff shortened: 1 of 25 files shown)


src/main/kotlin/app/hashers/FactHasher.kt
```diff
@@ -128,7 +127,8 @@ class FactHasher(private val serverRepo: Repo = Repo(),
 
         // Indentation.
         fsIndentation[email]!![FactCodes.INDENTATION_SPACES] +=
-            lines.count { it.isNotBlank()  &&  it.startsWith(" ") && !it.contains("\t")}
+            lines.count { it.isNotBlank()  &&  it.startsWith(" ") &&
```
yandzee: double space before `&&` ?

Issue #214, opened 2018-02-22
User is getting credit for somebody else's code


osilla:
feedback from user:

https://sourcerer.io/carlsson87
--

I have repos where I have committed entire frameworks, and I guess that all of that code is seen as “mine”, which leads to some weird conclusions (for example that I prefer tabs over spaces).

anatolystansler:
We add filtering of well-known vendor files: https://github.com/sourcerer-io/sourcerer-app/blob/develop/src/main/kotlin/app/hashers/Vendors.kt

Also, there is base support of stats-ignore config, but we need to update incremental update protocol to it start working properly. @asurkov let's discuss it offline.

PR #220, opened 2018-03-01
feat: support ignore paths (#214)

asurkov:


src/main/kotlin/app/hashers/CommitCrawler.kt
```diff
@@ -248,4 +269,34 @@ object CommitCrawler {
...
         }
     }
+
+    /**
+     * Return a list of paths that should be ignored in commit analysis.
+     */
+    private fun getIgnoredPaths(repo: Repository): List<String> {
+        return try {
+            val list = mutableListOf<String>()
+            val file = File(repo.workTree, CONF_FILE_PATH)
```
anatolykislov: We discussed that file structure and paths to vendor may change in time as well as config file. So it makes sense to get config from the currently processed commit. Suggested solution:
```
            var conf = null
            val treeWalk = TreeWalk(repo)
            treeWalk.reset(commitId)
            treeWalk.setFilter(PathFilter.create("./.sourcerer-conf"))
            if (treeWalk.next()) {
                val obj = repo.open(objectId)
                conf = RawText(obj.bytes)
            }
```

asurkov: oh, right, will fix it

src/main/kotlin/app/hashers/CommitCrawler.kt
```diff
@@ -156,6 +175,26 @@ object CommitCrawler {
...
             }
+            .filter { diff ->
+                val filePath =
+                    if (diff.getNewPath() != DiffEntry.DEV_NULL) {
+                        diff.getNewPath()
+                    } else {
+                        diff.getOldPath()
+                    }
+
+                // Update ignored paths list. The config file has retroactive
```
anatolykislov: Good thinking!

src/test/kotlin/test/tests/hashers/IgnorePathsTest.kt
```diff
@@ -0,0 +1,98 @@
+// Copyright 2017 Sourcerer Inc. All Rights Reserved.
+// Author: Liubov Yaronskaya (lyaronskaya@sourcerer.io)
```
anatolykislov: Fix authors and year.

src/main/kotlin/app/hashers/CommitCrawler.kt
```diff
@@ -156,6 +175,26 @@ object CommitCrawler {
...
+                // force, i.e. if it was added at this commit, then we presume
+                // it is applied to all commits, preceding this commit.
+                if (diff.getOldPath() == CONF_FILE_PATH) {
+                    ignoredPaths =
+                        getIgnoredPaths(repo, diff.getOldId().toObjectId())
+                }
+
+                !ignoredPaths.any { path ->
+                    filePath.startsWith(path)
```
anatolykislov: Conf contains "/test/a", file path "/test/apple.txt" and it will be filtered too. But we should have only filter /test/a file or dir.

src/main/kotlin/app/hashers/CommitCrawler.kt
```diff
@@ -156,6 +175,26 @@ object CommitCrawler {
...
+                    } else {
+                        diff.getOldPath()
+                    }
+
+                // Update ignored paths list. The config file has retroactive
+                // force, i.e. if it was added at this commit, then we presume
+                // it is applied to all commits, preceding this commit.
+                if (diff.getOldPath() == CONF_FILE_PATH) {
+                    ignoredPaths =
+                        getIgnoredPaths(repo, diff.getOldId().toObjectId())
```
anatolykislov: Should be getNewId().

asurkov: If this will be newId(), then the config file content will be equal to the cached one, so we don't update its content when rolling back in history, until I miss anything.

src/main/kotlin/app/hashers/CommitCrawler.kt
asurkov: right, will fix it

src/main/kotlin/app/hashers/CommitCrawler.kt
anatolykislov: It's ok that it may read current config twice when processing newest commit. We should use newId. 
For example: two commits in repo. Oldest commit contains config, but it is empty. Newest commit adds ignore paths to config. Initially before crawling we will read current config with ignore path. When crawling newest commit we will take old content of file, which is empty. So config will not be taken into account. Also, with old we will not be able to read config from oldest commit.

asurkov: So you think the config changes should follow right after the change or within the change, but it cann't go before the change. So for example
-> conf: ++test1
-> +test1/
-> conf: --test1 ++ test2
-> -test1/ + test2
ignored list is test2, so test2 folder is ignored, then ignored list is test2 (newId), and thus test1 folder is processed

which is weird, because it sort of breaks natural order of commits reading flow, but might be ok. We just should point out that config is applied to all preceding commits until overridden.

so going with newId then?

src/main/kotlin/app/hashers/CommitCrawler.kt
```diff
@@ -104,6 +111,18 @@ object CommitCrawler {
         df.setRepository(repo)
         df.setDetectRenames(true)
 
+        val confTreeWalk = TreeWalk(repo)
+        confTreeWalk.addTree(head.getTree())
+        confTreeWalk.setFilter(PathFilter.create(CONF_FILE_PATH))
+
+        var ignoredPaths =
+            if (confTreeWalk.next()) {
```
anatolykislov: I think this will read current repo HEAD, not default branch that we are selecting.

PR #225, opened 2018-03-05
feat: more imports processing

yaronskaya:


src/main/kotlin/app/extractors/CppExtractor.kt
```diff
@@ -32,12 +32,19 @@ class CppExtractor : ExtractorInterface {
...
-                val lineLib = res.groupValues.last()
+                val lineLib = res.groupValues
+                                 .last { !it.startsWith(".") && it != ""}
                 imports.add(lineLib)
             }
         }
-
         val libraries = imports.map { MULTI_IMPORT_TO_LIB.getOrDefault(it, it) }
+                               .map { import -> when {
+                                   import.startsWith("Q") -> "Qt"
```
anatolykislov: Every import that starts with "Q" will be Qt? Is it reasonable?

yaronskaya: At the moment yes

anatolykislov: Offline discussion

PR #272, opened 2018-04-22
Added some common javascript and .Net libraries

Thorium:


yaronskaya:
Hi, @Thorium. Thank you for the contribution. Actually, we use a bit deeper approach that is explained in [this repo](https://github.com/sourcerer-io/libraries). Basically, we need more information about libraries for linewise library detection. 
Could you please create a pull request there? If you have any questions, I'll be happy to answer.

PR #275, opened 2018-04-28
feat(tests): add lang heuristics testing

asurkov:
most of them are blind todos for now


asurkov:
I shouldn't have to put all those files within code changes. So here are the files you need to check:
src/test/kotlin/test/tests/extractors/HeuristicsTest.kt 
build.gradle
src/main/kotlin/app/extractors/Heuristics.kt

PR #283, opened 2018-06-01
feat: improve existing languages support

asurkov:


src/main/kotlin/app/extractors/Languages.kt
```diff
@@ -0,0 +1,108 @@
+// Copyright 2018 Sourcerer Inc. All Rights Reserved.
+// Author: Alexander Surkov (alex@sourcerer.io)
+
+package app.extractors
+
+object LG {
```
anatolykislov: image

anatolykislov: "Langs" would be more readable and short enough. It's also a good practice to name a file the same as a class inside.

asurkov: According to https://www.allacronyms.com/language/abbreviated language may be abbreviated to 'lg'. However if you prefer 'Lang' abbreviation, then 'Lang' look slightly better with me than its plural form.

PR #285, opened 2018-06-08
feat: add rust extractor

asurkov:


src/main/kotlin/app/extractors/RustExtractor.kt
```diff
@@ -0,0 +1,51 @@
...
+import app.model.CommitStats
+import app.model.DiffFile
+
+class RustExtractor : ExtractorInterface {
+    companion object {
+        const val LANGUAGE_NAME = Lang.Rust
+        val evaluator by lazy {
+            ExtractorInterface.getLibraryClassifier(LANGUAGE_NAME)
+        }
+        val importRegex = Regex("^extern crate \\w+;$")
```
yaronskaya: """ instead of single " allows not to use additional backslashes.

src/main/kotlin/app/extractors/RustExtractor.kt
```diff
@@ -0,0 +1,51 @@
...
+        val commentRegex = Regex("(////.+$)|(//[*].*?[*]//)")
```
yaronskaya: Are there four slashes in rust?!

asurkov: two as everywhere, just every backslash is backslashed, """ will help I think

yandzee: backslashed, not forwardslashed maybe?

asurkov: right, for some reason I "backslashed" them too :) we should have a test on this.

yandzee: wat :)

PR #287, opened 2018-06-18
feat: individual classifier for each library

yaronskaya:


src/main/kotlin/app/extractors/ClassifierManager.kt
```diff
@@ -42,7 +42,13 @@ class ClassifierManager {
             // Check line for usage of a library.
             val prediction = cache[libId]!!.evaluate(line)
             // Prediction based on two classes.
-            prediction[cache[libId]!!.libraries.indexOf(libId)] > 0.5
+            val prob = prediction[cache[libId]!!.libraries.indexOf(libId)]
+            // Libraries with no imports.
+            if (libId == "rb.rails") {
```
anatolystansler: Would be great to keep all language specific params in the proper extractors (e.g. all ruby heuristics should be in rubyextractor). This will help us with maintenance in future and improve readability for contributors. How can we achieve that?

PR #298, opened 2018-06-28
Allowed exts

asurkov:
* remove allowedExt thing - I guess it may improve performance slightly, but it breaks stuff having no extensions or non standard extensions, for example, XML
* copy vendors regexp from linguist


src/main/kotlin/app/hashers/CommitCrawler.kt
```diff
@@ -168,9 +168,11 @@ object CommitCrawler {
             }
             .filter { diff ->
                 val path = diff.newPath
-                val ext = FileHelper.getFileExtension(path)
-                if (allowedExts != null && !allowedExts.contains(ext)) {
-                    return@filter false
+                for (cnv in VendorConventions) {
+                    if (cnv.containsMatchIn(path)) {
+                        println("MATHCED: $path")
```
yaronskaya: MATCHED

asurkov: ah, debugging thing :)

anatolystansler:
@asurkov It's awesome that we will have vendor filter for hashing of repos. Good work!
We should also implement filtering for huge diffs, allowed extensions filtering helped with this problem partly, but it's gone now.

I'll fix all issues. Also, we using snake case for branch naming.

asurkov:
@anatolystansler, would you mind to file a new issue for huge diffs problem and cc me? thanks for fixing the issues btw.

asurkov:
btw, what's the point to follow name conventions for branches?

Issue #304, opened 2018-07-06
Support for Dart language


yaronskaya:
We've missed Dart language (referencing #291)

PR #312, opened 2018-07-14
feat: support dart language #304

asurkov:


src/main/kotlin/app/extractors/Languages.kt
```diff
@@ -18,6 +18,7 @@ object Lang {
     const val COMMONLISP = "lisp"  // Common Lisp
     const val CUDA = "cuda"  // Cuda
     const val D = "d"  // D
+    const val Dart = "dart"  // Dart
```
anatolystansler: We use UPPER CASE for constants according to styleguide.

src/main/kotlin/app/extractors/Extractors.kt
```diff
@@ -0,0 +1,41 @@
...
+        newLine = commentRegex.replace(newLine, "")
+        return super.tokenize(newLine)
+    }
+
+    override fun getLanguageName(): String? {
+        return language
+    }
+}
+
+object DartExtractor : ExtractorBase(
```
anatolystansler: Good idea!
Also, let's continue to store specific extractors in separate files.

asurkov:
mm, not sure what you mean. Do you want me to remove Dart extractor or
support Dart in awesome-libraries? It looks like a chicken-egg problem.

asurkov:
ah, it seems you have changed nice constant naming to less readable
uppercase names from the last time when I checked it :)

asurkov:
ok, it's good for discoverability, perhaps it makes sense to move them into
a separate folder then

yaronskaya:
@asurkov I mean calling CommonExtractor(Lang.DART).

PR #323, opened 2018-07-30
feat: support PLpgSQL module extraction

asurkov:


asurkov:
On Mon, Jul 30, 2018 at 8:20 AM, Anatoly Stansler <notifications@github.com>
wrote:

> *@anatolystansler* requested changes on this pull request.
>
> This PR has the same issue as #312
> <https://github.com/sourcerer-io/sourcerer-app/pull/312>. Import
> detection without classifiers will affect the performance of the app, but
> will not add any statistics for users.
> Also, an addition of PLpgSQL modules will not solve the problem of
> displaying a diversity of dialects of SQL in the profile.
>
It is the chicken egg problem. It won't take long to add some libraries,
will will improve statistics a bit of PLpgSQL users.


> Wonder maybe we should use ExtractorInterface for presented Extractors?
>
could you explain this more please?

PR #325, opened 2018-07-31
feat: support Coq language

asurkov:


src/main/kotlin/app/extractors/Heuristics.kt
```diff
@@ -1071,8 +1071,12 @@ val HeuristicsMap = mapOf<String, (String) -> ExtractorInterface?>(
     "ux" to { _ ->
         CommonExtractor(Lang.XML)
     },
-    "v" to { _ ->
-        CommonExtractor(Lang.VERILOG)
+    "v" to { buf ->
+        if (buf.contains("Require")) {
```
anatolystansler: Verilog files can contain "Require" too.

asurkov:
none in the repo's examples, you have a link at hands?

src/main/kotlin/app/extractors/Heuristics.kt
asurkov: @anatolystansler, also I see no 'Require' in keywords as well, http://www.xilinx.com/support/documentation/sw_manuals/xilinx11/ite_r_verilog_reserved_words.htm

anatolystansler: What about "Require" word in comments of verilog?

asurkov:
It definitely can give false positives. I guess it can be replaced on a
stricter regepx, which hopefully won't give false positives. Btw, the
concern is valid for other heuristics we have, and the issue is not so
common, since no complains and all tests pass.

anatolystansler:
We have 6 other heuristics that are based on `contains` function: 
- "<?hh"
- ":- module"
- "flowop"
- "last_client="
- "SOURCES"
- "HEADERS"

First 4 should be strict enough to not be in comments at least. Probably we can improve last 2, both of them related to QMAKE.

I think "Require" is too generic and probably used in comments often.

asurkov:
It deserves a new issue, please file one if agree. Note, I believe some
regexp should be adjusted as well, since some of them appear generic enough
to trigger false positives.

PR #338, opened 2018-08-29
feat(tech): add proba bound for plpgsql libs classifier.

yaronskaya:


src/main/kotlin/app/extractors/ClassifierManager.kt
```diff
@@ -50,7 +50,10 @@ class ClassifierManager {
             // Libraries with no imports.
             if (libId == "rb.rails") {
                 prob > 0.91
-            } else {
+            } else if (libId.startsWith(Lang.PLPGSQL)) {
+                prob > 0.7
```
asurkov: It looks good to the best of my knowledge. If you could add a comment above explaining what prob number is about, then it'd be great.

yaronskaya: Yep, added it.

PR #342, opened 2018-09-01
Matlab samples

asurkov:


asurkov:
#341 is based on this one, sorry for the confusion, but since you approved this one, I will land it then too. thanks!

PR #345, opened 2018-09-03
fix: do not extract import from comment in ExtractorBase, add Rails test.

yaronskaya:


yaronskaya:
The next PR will fix remaining languages. This one is just a quick fix of develop branch.

PR #346, opened 2018-09-11
use ExtractorBase for Swift

asurkov:


src/main/kotlin/app/extractors/SwiftExtractor.kt
```diff
@@ -4,35 +4,7 @@
...
-        return super.tokenize(newLine)
-    }
-
-    override fun getLanguageName(): String? {
-        return LANGUAGE_NAME
-    }
-}
+object SwiftExtractor : ExtractorBase(
+    language = Lang.SWIFT,
+    importRegex = Regex("""^(.*import)\s[^\n]*"""),
```
anatolystansler: `extractImportRegex` used for actually extracting of imports for choosing right classifiers
`importRegex` used to filter tokens from imports and as @yaronskaya works slightly faster

Here importRegex (less accurate version) used for both extracting and filter tokens. @yaronskaya review this change.

yaronskaya: @asurkov this regex doesn't extract imports. Please, use 'extractImportRegex' from previous version of SwiftExtractor. I wonder if it passes tests:)

asurkov:
@yaronskaya now I see what you say. The tests indeed pass, which means it has to be a problem with the tests. For example, there's one `val line = "import Foundation"`, which is supposed to extract no libraries. Do you know whether this is correct one? Is there any special about Swift import instructions?

yaronskaya:
@asurkov it turned out that we don't have tests of imports for Swift, there are only tests for libraries extraction. Could you please add them?

src/test/kotlin/test/tests/extractors/ExtractorTest.kt
```diff
@@ -368,4 +357,33 @@ class ExtractorTest : Spek({
...
+            var lines = listOf("/* class City: RLMObject { */")
+            var actualLineImports = SwiftExtractor.extractImports(lines)
+            assertTrue(actualLineImports.isEmpty())
+        }
+        it("Swift imports") {
+            var lines = listOf(
+                "import UIKit"
+            )
+            var actualLineImports = SwiftExtractor.extractImports(lines)
+            assertEquals(actualLineImports, listOf("UIKit"))
```
yaronskaya: Use `assertExtractsImport`

PR #376, opened 2018-10-23
feat: colleagues(fast)

yaronskaya:


src/main/kotlin/app/hashers/CommitCrawler.kt
```diff
@@ -156,9 +167,9 @@ object CommitCrawler {
                 commit = parentCommit
                 continue
```
anatolystansler: Paths calculation will be affected

2Blessed2BstressedBYtheevilOnesMess2018:
Pull

PR #378, opened 2018-10-24
feat: support scala import extraction

asurkov:


src/main/kotlin/app/extractors/ScalaExtractor.kt
```diff
@@ -0,0 +1,9 @@
+// Copyright 2018 Sourcerer Inc. All Rights Reserved.
+// Author: Alexander Surkov (alex@sourcerer.io)
+
+package app.extractors
+
+object ScalaExtractor : ExtractorBase(
+    language = Lang.SCALA,
+    importRegex = Regex("""^import (?:_root_\.)?(\w+)"""),
```
yaronskaya: The regex doesn't extract import with two or more words in it like in org.scalatest.

Issue #121, opened 2017-12-08
(Fun Fact) I use template metaprogramming.


osilla:
I use template metaprogramming.

PR #386, opened 2018-10-29
feat: cpp template metaprogramming fun fact (#121)

yaronskaya:


src/main/kotlin/app/extractors/CppExtractor.kt
```diff
@@ -40,4 +64,8 @@ class CppExtractor : ExtractorInterface {
     override fun getLanguageName(): String? {
         return LANGUAGE_NAME
     }
+
+    private fun isTemplate(line: String): Boolean {
+        return line.contains("template <")
```
anatolystansler: What about multiline templates?

anatolystansler: Should be ok for first version, but we should look for "template<" too.

Issue #308, opened 2018-07-10
Add several repo with add command


mathieutu:
Hi, 
It should be cool to be allowed to add several repo directly with the add command.
For now, we're forced to "add" them one by one.

And if allowed it will be possible to use wildcard too.

```zsh
$ sourcerer add foo bar 
[i] App started.
[e] Uncaught exception: Only one main parameter allowed but found several: "foo" and "bar"

$ sourcerer add projects/*
[i] App started.
[e] Uncaught exception: Only one main parameter allowed but found several: "projects/foo" and "projects/bar"
```

Thanks for your amazing work!

anatolystansler:
Hi @mathieutu 

This is a really good suggestion! We plan to update the app interface in the future and will take this into account.

sfat:
@yaronskaya we can close this, right?

yaronskaya:
@sfat sure

PR #535, opened 2019-08-17
feat: Support multiple repositories specified from the same line (#308)

sfat:
Added support for issue #308


sfat:
I could only knew how to test going with the route of reaching to `AddRepoState.kt`.
It's not very clear how to make the `sourcerer` cli built on my local environment.
The ticket is talking about running the command `sourcerer add 1 2 3`.
I only did it via `java -jar build/libs/sourcerer.jar` which seems to be going on a different flow.

PR #538, opened 2019-08-23
feat: support quasar

asurkov:


yaronskaya:
@asurkov we also need to add it to awesome-libraries.

